PR #5, opened 2022-02-04
'Solution'

serhii56465:


denyslipin (approved):

tests/test_main.py
```diff
@@ -1 +1,34 @@
-# write your code here
+from unittest import mock
+from app.main import can_access_google_page, valid_google_url, has_internet_connection
+
+
+@mock.patch("app.main.valid_google_url")
+@mock.patch("app.main.has_internet_connection")
+def test_url_true_connection_true(internet_connection, valid_url):
```
MasterpieceElbow, line 7: It is bad test names. It would be better to write `valid/correct url` and `has connection / connection exists`.
Rewrite all test names.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #12, opened 2022-02-05
'Solution'

Oleksandr563487953:


serhii56465 (approved):
In my opinion, this task doesn't need to parameterize. Without params, test code looks more clear, easier for this task. But it's just my from my point of view :)

Oleksandr563487953:
> In my opinion, this task doesn't need to parameterize. Without params, test code looks more clear, easier for this task. But it's just my from my point of view :)

First version was provided without parametrize, but i observed that there was no tests for different situations, just mocked function.

PR #20, opened 2022-02-07
Solution

Geksus:


tests/test_main.py
```diff
@@ -1 +1,51 @@
...
+@mock.patch("app.main.valid_google_url")
+@mock.patch("app.main.has_internet_connection")
+def test_accessible_time_window(
+        mocked_valid_google_url,
+        mocked_internet_connection
+):
+    mocked_valid_google_url.return_value = True
+    mocked_internet_connection.return_value = False
+
+    assert can_access_google_page("https://google.com") == "Not accessible"
```
serhii56465, line 51: It seems like you repeated your previous test case.

Geksus: Thank's. Fixed

serhii56465 (approved):

tests/test_main.py
```diff
@@ -1 +1,39 @@
...
+@mock.patch("app.main.valid_google_url")
+@mock.patch("app.main.has_internet_connection")
+def test_if_connection_is_off_and_url_is_correct(
+        mocked_valid_google_url,
+        mocked_internet_connection
+):
+    mocked_valid_google_url.return_value = True
+    mocked_internet_connection.return_value = False
+
+    assert can_access_google_page("https://google.com") == "Not accessible"
```
MasterpieceElbow, line 39: What if both `valid_google_url` and `internet_connection` return `False`?

MasterpieceElbow (changes requested):

tests/test_main.py
Geksus: Added another test

MasterpieceElbow (approved):

PR #107, opened 2022-09-23
solution

yevhenii-nevmyvako:


app/test_main.py
```diff
@@ -1 +1,50 @@
...
+           "Not accessible"
+
+
+@mock.patch('app.main.valid_google_url')
+@mock.patch('app.main.has_internet_connection')
+def test_return_access_denied_if_time_and_url_wrong(
+        mocked_valid_google_url,
+        mocked_has_internet_connect
+):
+    mocked_valid_google_url.return_value = True
```
IvanRamyk, line 47: duplicated previous test

IvanRamyk (changes requested):

app/test_main.py
yevhenii-nevmyvako: remake

DmytroSvirsa (approved):

PR #114, opened 2022-09-28
Solution

AVyha:


mltdll (approved):

app/test_main.py
```diff
@@ -1 +1,42 @@
...
+@pytest.fixture()
+def mocked_has_internet_connection():
+    with mock.patch("app.main.has_internet_connection") as mock_has_internet:
+        yield mock_has_internet
+
+
+def test_function_valid_google_url_is_correct(mocked_has_internet_connection):
+    mocked_has_internet_connection.return_value = True
+
+    assert can_access_google_page("http://www.google.com/") == "Accessible"
```
Nattalli, line 21: Add an error message. It's good when user understands what is the problem

app/test_main.py
```diff
@@ -1 +1,42 @@
...
+
+
+def test_should_return_accessible(mocked_valid_url,
+                                  mocked_has_internet_connection):
+    mocked_valid_url.return_value = True
+    mocked_has_internet_connection.return_value = True
+    assert can_access_google_page("") == "Accessible"
```
Nattalli, line 28: Add an error message. It's good when user understands what is the problem

app/test_main.py
```diff
@@ -1 +1,42 @@
...
+    mocked_valid_url.return_value = False
+    mocked_has_internet_connection.return_value = True
+    assert can_access_google_page("") == "Not accessible"
+
+
+def test_if_only_valid_url_true(mocked_valid_url,
+                                mocked_has_internet_connection):
+    mocked_valid_url.return_value = True
+    mocked_has_internet_connection.return_value = False
+    assert can_access_google_page("") == "Not accessible"
```
Nattalli, line 42: додай тест, де дві умови дорівнюють `False`

app/test_main.py
```diff
@@ -1 +1,42 @@
...
+    mocked_valid_url.return_value = True
+    mocked_has_internet_connection.return_value = True
+    assert can_access_google_page("") == "Accessible"
+
+
+def test_if_only_internet_connection_true(mocked_valid_url,
+                                          mocked_has_internet_connection):
+    mocked_valid_url.return_value = False
+    mocked_has_internet_connection.return_value = True
+    assert can_access_google_page("") == "Not accessible"
```
Nattalli, line 35: Add an error message. It's good when user understands what is the problem

app/test_main.py
Nattalli, line 42: Add an error message. It's good when user understands what is the problem

app/test_main.py
Nattalli, line 21: No, this test is incorrect. The condition is `returns "Accessible" if url is valid to access the Google home page and it has internet connection`. So both are required

Nattalli (changes requested):

app/test_main.py
```diff
@@ -1 +1,59 @@
...
+    assert can_access_google_page("") == "Accessible", \
+        "Function must return 'Accessible' when " \
+        "internet connection and google url return True"
+
+
+def test_function_valid_google_url_is_correct(mocked_has_internet_connection):
+    mocked_has_internet_connection.return_value = True
+
+    assert can_access_google_page("http://www.google.com/") == "Accessible", \
+        "Function must return 'Accessible' if url is correct"
```
Nattalli, line 32: You can remove this test as the function must return `True` if only both `mocked_valid_url` and `mocked_has_internet_connection` return True

Nattalli (changes requested):

Nattalli (approved):
Good job! 😸

PR #119, opened 2022-09-30
Solution

kateryna3k:


app/test_main.py
```diff
@@ -1 +1,43 @@
...
+@pytest.fixture()
+def mocked_has_internet_connection():
+    with mock.patch("app.main.has_internet_connection") as mock_has_internet:
+        yield mock_has_internet
+
+
+def test_all_works(mocked_valid_url, mocked_has_internet_connection):
+    mocked_valid_url.return_value = True
+    mocked_has_internet_connection.return_value = True
+    assert can_access_google_page("www.google.com/") == "Accessible", \
```
mltdll, line 21: Sorry for bringing it up again, but now I have an actual quote from our mentor Oleksii to reference:
> Avoid using `\`. Use `( )`

mltdll (changes requested):

app/test_main.py
kateryna3k: thx for your review, I've fixed it recently :))

mltdll (approved):

TPolina (approved):
Well done!

PR #121, opened 2022-10-01
Solution

taras-andruschenko:


app/test_main.py
```diff
@@ -1 +1,49 @@
...
+from app.main import can_access_google_page
+
+
+@mock.patch("app.main.valid_google_url")
+@mock.patch("app.main.has_internet_connection")
+def test_can_access_page(mocked_valid_url, mocked_internet) -> None:
+    mocked_valid_url.return_value = True
+    mocked_internet.return_value = True
+
+    assert can_access_google_page("some url") == "Accessible", \
```
mltdll, line 12: You should use `()` instead of `\`. For other assertions as well.

mltdll (changes requested):

mltdll (approved):

TPolina (approved):
Well done!

PR #122, opened 2022-10-02
Solution

lksndr307:


app/test_main.py
```diff
@@ -1 +1,28 @@
-# write your code here
+from unittest import mock
+
+import pytest
+
+from app.main import can_access_google_page
+
+
+class TestAccessGooglePage:
```
Abnormaltype, line 8: if you use `pytest`, you should write function-based tests

Abnormaltype (changes requested):

Abnormaltype (approved):
Well done!

PR #124, opened 2022-10-02
'Solution'

TetyanaPavlyuk:


app/test_main.py
```diff
@@ -1 +1,35 @@
-# write your code here
+import pytest
+from unittest import mock
+from app.main import can_access_google_page
+
+
+class TestCanAccessGooglePage:
```
Abnormaltype, line 6: if you use `pytest` you should write function-based tests

Abnormaltype (changes requested):

Abnormaltype (approved):
Well done!

PR #128, opened 2022-10-04
Solution Vadim Sabadyr

VadimSabadyr:


app/test_main.py
```diff
@@ -1 +1,28 @@
...
+def test_google_page_true(mocked_google_url, mocked_internet_connection):
+    mocked_google_url.return_value = True
+    mocked_internet_connection.return_value = True
+    assert can_access_google_page(100) == "Accessible"
+
+
+def test_google_page_false(mocked_google_url, mocked_internet_connection):
+    mocked_google_url.return_value = False
+    mocked_internet_connection.return_value = True
+    assert can_access_google_page(400) == "Not accessible"
```
Gr1zzZzzly, line 28: i think u must check all variants of True and False
True , False
False, False
And why are u using int in "can_access_google_page(400)" it's doesn't matter but little confused

Gr1zzZzzly (changes requested):

app/test_main.py
Y-Havryliv: I agree with Igor, in tests we should cover all possible variants, it works now with your solution, but what if `can_access_google_page` was written with mistakes?

Y-Havryliv (changes requested):

Y-Havryliv (approved):

PR #146, opened 2022-10-06
Solution

Kitazuka:


TPolina (changes requested):
one test is missing. What if invalid url and no internet?

PR #178, opened 2022-10-24
'Solution'

rakamakaphone:


app/main.py
```diff
@@ -13,7 +13,7 @@ def has_internet_connection() -> bool:
 
 
 def can_access_google_page(url: str) -> str:
-    if has_internet_connection() and valid_google_url(url):
+    if valid_google_url(url) and has_internet_connection():
```
Abnormaltype, line 16: why did you do this?

Abnormaltype (changes requested):

Abnormaltype (approved):
Well done!

Abnormaltype (approved):

app/main.py
rakamakaphone: better don't ask :::)))
I didn't know how to mock two func same time. So I tried different ways, than returned back, but not the same order.

PR #182, opened 2022-10-28
Solution to google page

GeorgePavlej:


app/test_main.py
```diff
@@ -1 +1,66 @@
...
+
+@pytest.fixture()
+def mock_internet_connection() -> None:
+    with mock.patch("app.main.has_internet_connection") \
+            as mock_connection:
+        yield mock_connection
+
+
+def test_true_if_valid_url_and_has_connection(
+        mock_valid_google_url: Callable,
```
Serhii-Leonenko, line 22: change annotation `mock` instead of `Callable` also in all similar cases

Serhii-Leonenko (changes requested):

IvanRamyk (approved):

PR #196, opened 2022-11-30
Solution

MykytaMoshchenko:


IvanRamyk (changes requested):
please, consider all possible combinations

app/test_main.py
```diff
@@ -1 +1,57 @@
...
+
+
+def test_accessible_when_internet_and_url_are_not_ok(
+        mock_internet_connection: mock,
+        mock_valid_url: mock
+) -> None:
+    mock_internet_connection.return_value = True
+    mock_valid_url.return_value = True
+    assert can_access_google_page(
+        "https://www.wikipedia.org/") == "Accessible"
```
Abnormaltype, line 57: should be both `False`?

Abnormaltype (changes requested):

app/test_main.py
MykytaMoshchenko: Yes, should be both False.
Corrected.

Abnormaltype (approved):

PR #197, opened 2022-11-30
Solution

Vladyslav7125:


app/test_main.py
```diff
@@ -1 +1,43 @@
-# write your code here
+# git commit -am 'Solution'
```
Serhii-Leonenko, line 1: remove unnecessary comment

app/test_main.py
```diff
@@ -1 +1,43 @@
...
+from unittest import mock
+from app.main import can_access_google_page
```
Serhii-Leonenko, line 3: you should split this imports with empty line

Serhii-Leonenko (changes requested):

app/test_main.py
Vladyslav7125: Fixed

Vladyslav7125: What about my task?

Vladyslav7125: What about my task?

IvanRamyk (approved):
Well done!

PR #198, opened 2022-11-30
'Solution'

Mazur-Mariia-Python:


OleksandrPozden (changes requested):

IvanRamyk (approved):
also it'd be nice to check both False values but let it be)

PR #200, opened 2022-12-07
Solution Maksym Maksimov

MKeSiMu:


Abnormaltype (changes requested):
check when has no internet and no valid URL

YuriiKindrat (approved):

PR #205, opened 2022-12-14
Solution

Oleksiy-Liubchenko:


Oleksiy-Liubchenko:
Am I right, in test where I mocked two functions, I can't understand and test is link in function can_access_gooogle_page() right, yes? And in last test I just need to put  incorrect link as I did?

Serhii-Leonenko (approved):
It doesn't matter which link you will use in your tests. Because all function depends on functions which you mocked.

PR #206, opened 2022-12-14
Solution

Nexxel05:


Serhii-Leonenko (changes requested):
Add one more test when both are False

IvanRamyk (approved):

PR #216, opened 2022-12-31
Solution

andriy-md:


app/test_main.py
```diff
@@ -1 +1,27 @@
...
+@pytest.mark.parametrize(
+    "valid_url,internet_connection,expected",
+    [
+        (True, True, "Accessible"),
+        (False, True, "Not accessible"),
+        (True, False, "Not accessible"),
+        (False, False, "Not accessible"),
+    ]
+)
+def test_can_access_google_page_if_internet_connection_and_valid_url(
```
Vasyl-Smutok, line 16: I think you better split it into several tests

Vasyl-Smutok (changes requested):

app/test_main.py
andriy-md: Hello!
I tried to think how to split this into different tests, but nothing came to mind.
To test if function `can_access_google_page` returns correct value, I need to mock both functions. If I don't test would not work properly.

Please, give some extra hints)

Abnormaltype: you can use `id` in `parametrize` to describe your test

Abnormaltype (changes requested):

app/test_main.py
```diff
@@ -1 +1,43 @@
...
+
+
+@pytest.mark.parametrize(
+    "valid_url,internet_connection,expected",
+    [
+        pytest.param(
+            True,
+            True,
+            "Accessible",
+            id="URL is valid, has internet connection"),
```
katevasilenko, line 14:
```suggestion
            id="URL is valid, has internet connection"
        ),
```
It’s picky but could you provide this code style all over the test data

katevasilenko (changes requested):

app/test_main.py
```diff
@@ -1 +1,43 @@
...
+        pytest.param(
+            False,
+            True,
+            "Not accessible",
+            id="URL not valid, has internet connection"),
```
katevasilenko, line 19:
```suggestion
            id="URL not valid, has internet connection"
        ),
```

app/test_main.py
```diff
@@ -1 +1,43 @@
...
+        pytest.param(
+            True,
+            False,
+            "Not accessible",
+            id="URL is valid, no internet connection"),
+        pytest.param(
+            False,
+            False,
+            "Not accessible",
+            id="URL not valid, no internet connection"),
```
katevasilenko, line 29:
```suggestion
            id="URL not valid, no internet connection"
        ),
```

ArsenPidhoretskyi (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #218, opened 2023-01-04
tests

VItalik1292:


Abnormaltype (changes requested):
you should write 4 cases:
when
Internet connection is True
URL is correct

internet connection is True
The URL is not correct 

and vice versa

PR #219, opened 2023-01-04
Validating url and connection

MarynaProkhorenko:


Abnormaltype (changes requested):
what about when are both True or False?

Abnormaltype (approved):
Well done!

PR #220, opened 2023-01-06
Solution

Tania-Kharchuk:


Tania-Kharchuk:
Could you please tell me, are mocked_has_internet_connection and mocked_valid_google_url callable or bool or smth else?

katevasilenko (changes requested):
Please write your test using `pytest.mark.parametrize` Now your tests are self repeating.
Also answering your question, I suppose `bool` is ok

PR #226, opened 2023-01-09
Develop

roma-mechanic:


app/test_main.py
```diff
@@ -1 +1,48 @@
...
+
+
+def test_can_access_google_page_if_url_true(
+        mocked_valid_google_url: object,
+        mocked_has_internet_connection: object
+) -> None:
+    test_url = "https://test_url"
+    mocked_valid_google_url.return_value = True
+    mocked_has_internet_connection.return_value = False
+    assert can_access_google_page(test_url) == "Accessible"
```
YuriiKindrat, line 38: here u have an valid url but has no internet connection 
image

YuriiKindrat: Also add the test for case with internet connection but without valid url

YuriiKindrat (changes requested):

andrewbelo (approved):

PR #245, opened 2023-02-03
Solution

YehorKorn:


app/main.py
```diff
@@ -17,3 +17,9 @@ def can_access_google_page(url: str) -> str:
         return "Accessible"
     else:
         return "Not accessible"
+
+
+# response = requests.get("https://www.google.com/")
+# print(response.status_code)
+#
+# print(datetime.time(hour=23))
```
ArsenPidhoretskyi, line 25: Remove this

app/test_main.py
```diff
@@ -1 +1,44 @@
...
+    with (mock.patch("app.main.has_internet_connection")
+          as mocked_has_internet_connection):
+        yield mocked_has_internet_connection
+
+
+def test_should_return_accessible_if_all_checks_are_valid(
+        mock_has_internet_connection: mock.MagicMock,
+) -> None:
+    mock_has_internet_connection.return_value = True
+    assert can_access_google_page("https://www.google.com/") == "Accessible"
```
ArsenPidhoretskyi, line 24: Also, mock here mock_valid_google_url. Do not need to send really url) For example, if google stops working your code will fail)

ArsenPidhoretskyi (changes requested):
Use @pytest.parametrize and check 4 possible variants

Vasyl-Smutok (approved):
Good Job 👍

PR #246, opened 2023-02-04
'Solution'

Paul-Maslov:


app/test_main.py
```diff
@@ -1 +1,48 @@
...
+
+
+def test_should_not_get_google_page_if_no_page(
+        mocked_has_internet_connection: bool,
+        mocked_valid_google_url: bool
+) -> None:
+    mocked_has_internet_connection.return_value = True
+    mocked_valid_google_url.return_value = False
+    assert can_access_google_page("https://www.google.com/")\
+           == "Not accessible"
```
ArsenPidhoretskyi, line 48: Use @pytest.parametrize and check all variants(4)

ArsenPidhoretskyi (changes requested):

Bohdan-Salamakha (approved):

app/test_main.py
Paul-Maslov: Unfortunately, mock inside parametrize does not work, and I did not find an example in the documentation

Bohdan-Salamakha (approved):

PR #253, opened 2023-02-11
solution

mkobylianskyi:


app/test_main.py
```diff
@@ -1 +1,57 @@
-# write your code here
+from typing import Callable
+from unittest import mock
+from app.main import can_access_google_page
+
+import pytest
```
katevasilenko, line 5: Pytest should be in first block of imports, than blank line and than your own import

app/test_main.py
```diff
@@ -1 +1,57 @@
...
+
+
+@pytest.fixture
```
katevasilenko, line 8:
```suggestion
@pytest.fixture()
```

app/test_main.py
```diff
@@ -1 +1,57 @@
...
+def test_valid_google_url() -> None:
```
katevasilenko, line 9: Please don’t start name of this function with `test`. It’s better to use `mock`

app/test_main.py
```diff
@@ -1 +1,57 @@
...
+    with mock.patch("app.main.valid_google_url") as mocked_url:
+        yield mocked_url
+
+
+@pytest.fixture
+def test_has_internet_connection() -> None:
```
katevasilenko, line 15: The same as in previous mock

app/test_main.py
```diff
@@ -1 +1,57 @@
...
+    with mock.patch("app.main.has_internet_connection") as mocked_connection:
+        yield mocked_connection
+
+
+def test_incorrect_url(
```
katevasilenko, line 20: I would prefer if you represent internet connection behaviour in test name

app/test_main.py
```diff
@@ -1 +1,57 @@
...
+        test_valid_google_url: Callable,
+        test_has_internet_connection: Callable
+) -> None:
+    test_valid_google_url.return_value = False
+    test_has_internet_connection.return_value = True
+
+    assert can_access_google_page("www.gog.co") == "Not accessible"
+
+
+def test_no_internet_connection(
```
katevasilenko, line 30: And here the same about representing url validation in test name

katevasilenko (changes requested):

Vasyl-Smutok (approved):
Well Done 👍

PR #254, opened 2023-02-13
Solution

SlavikSherbak:


app/test_main.py
```diff
@@ -1 +1,58 @@
...
+    assert can_access_google_page("www.googl.com") == "Not accessible"
+
+
+def test_valid_googl_url_and_have_internet_connection(
+        mock_google_url: Callable,
+        mock_internet_connection: Callable
+) -> None:
+    mock_google_url.return_value = True
+    mock_internet_connection.return_value = True
+    assert can_access_google_page("www.googl.com") == "Accessible"
```
Grayder0152, line 58: Use `pytest.mark.parametrize` instead of this code

Grayder0152 (changes requested):

ArsenPidhoretskyi (changes requested):
Use pytest.parametrize. Because it is the same tests but a litle changes

ArsenPidhoretskyi (approved):
Well done)

PR #257, opened 2023-02-15
Solution

manjustice:


AndereLion (approved):
Super

app/test_main.py
```diff
@@ -1 +1,29 @@
...
+
+from .main import can_access_google_page
+
+
+@pytest.mark.parametrize(
+    "url,status,internet,result",
+    [
+        ("google.com", True, False, "Not accessible"),
+        ("gooogle.com", False, True, "Not accessible"),
+        ("google.com", True, True, "Accessible")
```
YuriiKindrat, line 13: no need to repeat url for 3 times, u can create this variable inside ur test only once

YuriiKindrat: also what about false/false condition ?

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #259, opened 2023-02-15
1 commit

AndereLion:


SafonovVladimir (approved):
GJ

YuriiKindrat (approved):

app/test_main.py
```diff
@@ -1 +1,27 @@
-# write your code here
+from app.main import can_access_google_page
+
+from unittest import mock
+import pytest
```
YuriiKindrat, line 4: https://peps.python.org/pep-0008/#:~:text=Imports%20are%20always,library%20specific%20imports.
fix this

YuriiKindrat (changes requested):

app/test_main.py
AndereLion: Fixed

YuriiKindrat (approved):

PR #262, opened 2023-02-17
Solution

AlimFreight:


app/test_main.py
```diff
@@ -1 +1,30 @@
-# write your code here
+import pytest
+
+from unittest import mock
+
+from app.main import can_access_google_page
+
+
+@pytest.mark.parametrize(
+    "allowed_url, allowed_time, result",
```
vasylhnatiuk, line 9: would be a better name `internet_connection` instead of `allowed_time`

vasylhnatiuk (approved):

PR #264, opened 2023-02-18
Develop

MaybeTI:


WDemidenko (approved):
Good job!

app/test_main.py
```diff
@@ -1 +1,40 @@
...
+import pytest
+
+
+@pytest.mark.parametrize(
+    "url,valid_url,connection,result",
+    [
+        ("https://www.google.com.ua", True, True, "Accessible"),
+        ("https://www.google.com.ua", False, True, "Not accessible"),
+        ("https://www.google.com.ua", True, False, "Not accessible"),
+        ("https://www.google.com.ua", False, False, "Not accessible")
```
WDemidenko, line 14: It would be better to remove url from here, and paste it once in can_access_google_page("https://www.google.com.ua")

YuriiKindrat: @WDemidenko is right, make ur code DRY

YuriiKindrat (changes requested):

app/test_main.py
```diff
@@ -1 +1,39 @@
-# write your code here
+from typing import Callable
+from unittest import mock
+from app.main import can_access_google_page
+
+import pytest
```
YuriiKindrat, line 5: fix this, according to best practicies 
https://peps.python.org/pep-0008/#:~:text=Imports%20are%20always,group%20of%20imports.

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #267, opened 2023-02-20
Solution

WDemidenko:


app/test_main.py
```diff
@@ -1 +1,51 @@
...
+        pytest.param(
+            True, True, "Accessible",
+            id="if valid url and internet connection are both True"
+        )
+    ]
+)
+def test_for_some(
+        valid_url: bool,
+        internet_connection: bool,
+        can_access: bool
```
NEkropo1, line 27: you can make these names a little bit more descriptive. For example, you can name you func test_can_access_google_page and change the parameter names to `is_valid_url, has_internet_connection, expected_result`
But it's ok, only func name is a little pain.

app/test_main.py
```diff
@@ -1 +1,51 @@
...
+
+
+def test_internet_connection_and_valid_url_was_called() -> None:
+    with mock.patch("app.main.valid_google_url") as mocked_valid_url:
+        with mock.patch(
+                "app.main.has_internet_connection"
+        ) as mocked_internet_connection:
+            can_access_google_page("www.ua")
+            mocked_internet_connection.assert_called_once()
+            mocked_valid_url.assert_called_once()
```
NEkropo1, line 51: you could simplify code by using `mock.patch.object` instead of using the `with` statements, for example:
```
with mock.patch.object(
            app.main, "valid_google_url"
    ) as mocked_valid_url, mock.patch.object(
            app.main, "has_internet_connection"
    ) as mocked_internet_connection:
        can_access_google_page("www.ua")
        mocked_valid_url.assert_called_once()
        mocked_internet_connection.assert_called_once()
```
or simply use decorators )

NEkropo1 (changes requested):

app/test_main.py
WDemidenko: Forget to change test function name)

YuriiKindrat (changes requested):
What about False/False condition ?

YuriiKindrat (approved):

PR #268, opened 2023-02-20
Solution

NEkropo1:


WDemidenko (approved):
It would be better to create pytest.mark.parametrize to check can_access function works correctly in all cases, and also a separate test to check that both internal functions are called inside can_access function. Because for me the code looks not quite clean, with unnecessary repetitions. But its up to you

NEkropo1:
> It would be better to create pytest.mark.parametrize to check can_access function works correctly in all cases, and also a separate test to check that both internal functions are called inside can_access function. Because for me the code looks not quite clean, with unnecessary repetitions. But its up to you

As for me, test idea is just in covering most situations, which could be covered. I mean, in exact test case, there're only 4 variation of output, and since I test once in 2 different tests their summoning(idk, `callage`?), it should work for all other cases too. Same as if I provided bonus test to check different calls.

app/test_main.py
```diff
@@ -1 +1,43 @@
...
+
+
+@mock.patch("app.main.has_internet_connection")
+@mock.patch("app.main.valid_google_url")
+def test_can_access_google_page(mocked_valid_google_url: mock,
+                                mocked_has_connection: mock) -> None:
+    mocked_has_connection.return_value = True
+    mocked_valid_google_url.return_value = True
+    assert can_access_google_page(URL) == "Accessible"
+    assert mocked_valid_google_url.called_once_with(URL)
```
YuriiKindrat, line 31: why do u need this test? u have already tested this condition in first test

YuriiKindrat: it is better to add False/False condition here

YuriiKindrat (changes requested):

app/test_main.py
NEkropo1: I disagree, test coverage is 100% of basic tests: 
At first i check if time is ok,  
at second if time is non ok, and func_called_once
at last if url is not valid and time is ok
in this if both connection is ok and time is ok + url_check_called_once_with_same_input
So it covers every single event possible, even False/False
Same, as python works with `__ge__` `__eq__` if you got my point.

NEkropo1: Furthermore if I change this test_case to False False, it won't cover True from url_check if time_check is False

Or I'm missing something

NEkropo1: The only reason I could at False False is simplicity and explicit, but since it's tests by pytest, and not main code, it changes nothing, 'cause provided info from traceback will cover all cases in just enough way

YuriiKindrat: first test have valid url so valid url fucn will return - True
u will test the case for True/True condition
next test is the same but with True/False condition
then u are checking True/True condition in 3d test,
and in 4th test u are check for False/True condition 
Where did u check a False/False condition ?

YuriiKindrat (changes requested):

app/test_main.py
NEkropo1: I apologize if my previous explanation was not clear enough. As I mentioned earlier, when I tested the scenarios with False/True and True/False, it also covers the scenario of False/False. I hope this clarifies any confusion you may have had. To further illustrate, consider the example of giving you one lasagna and one apple. If someone were to ask you separately if you received the lasagna and the apple, it would cover both cases.
I could reorganize an idea as a list of possible situations and parsing, while showing on which test it's full, and on which it's part, but it's pretty simple, even gpt suggests to add False/False ONLY if it's not clear enough in code, but this code is simple permutations of 1, 2 states, idk.

Serhii-Leonenko: Imagine that it's not your own code and you even don't see implementation at all, just know the names of the functions and your task is to write tests that will cover all cases(true/true, false/false, true/false, false/true)

Serhii-Leonenko (changes requested):

app/test_main.py
NEkropo1: I'm not QA for writing tests for others) Furthermore, if this case gonna be a hard nut I gonna ask contacts of dev who made this and ask for fast brief before making tests(cause if some functionality is overpowered and you got tight time condition it would help much more, than writing tests). And If something happened, for example he's fired for that code or we need to know which problems it could handle, I guess the correct decision will be writing code myself and I still won't need these tests, or simply concede and stop wasting time on useless ideas, because proper time, product management and maintenance are crucial for every project purpose.

app/test_main.py
```diff
@@ -1 +1,50 @@
...
+            False,
+            "Not accessible",
+        ),
+        (
+            "https://www.google.com",
+            False,
+            False,
+            "Not accessible",
+        ),
+    ],
```
YuriiKindrat, line 35: no need to write url here as param, u can pass it directly to can_access function

app/test_main.py
```diff
@@ -1 +1,50 @@
...
+        "invalid url",
+        "no internet and invalid url",
+    ]
+)
+def test_can_access_google_page(url: str,
+                                mock_internet: bool,
+                                mock_valid: bool,
+                                expected: str) -> None:
+    with mock.patch("app.main.valid_google_url", return_value=mock_valid):
+        with mock.patch("app.main.has_internet_connection",
```
YuriiKindrat, line 48: no need to use "with" keyword twice here

YuriiKindrat (changes requested):

Serhii-Leonenko (approved):

PR #269, opened 2023-02-21
Solution

ZabFTFT:


DavidGlusch (approved):

app/test_main.py
```diff
@@ -1 +1,38 @@
...
+        pytest.param(
+            False,
+            True,
+            "Not accessible",
+            id="should not get access with broken url"),
+        pytest.param(
+            True,
+            False,
+            "Not accessible",
+            id="should not get access without internet connection "),
```
YuriiKindrat, line 24: fix indentation for pytest.param

YuriiKindrat (changes requested):

SergiiMachulin (approved):

YuriiKindrat (changes requested):
what about False/False condition ?

ZabFTFT:
> what about False/False condition ?

I thought we don't need it because we have logic operator AND inside tested function so no matter what if one of conditions is False,result will be the same as double False.But I fixed it. =)

oksana-feshchenko (approved):

Serhii-Leonenko (approved):

PR #270, opened 2023-02-21
solution

DavidGlusch:


Maytreyaya (approved):

app/test_main.py
```diff
@@ -1 +1,32 @@
...
+from app.main import can_access_google_page
+
+
+@pytest.mark.parametrize(
+    "url, valid_google_url, has_internet_connection, result",
+    [
+        ("google.com", True, True, "Accessible"),
+        ("1234.com", False, True, "Not accessible"),
+        ("google.com", True, False, "Not accessible"),
+        ("5678.com", False, False, "Not accessible")
```
YuriiKindrat, line 12: no need to use different urls for this tests, so u can write url only once inside ur test

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #272, opened 2023-02-22
added tests

drapail:


ArsenPidhoretskyi (changes requested):
Use:
from unittest import mock no need monkeypatch
Also, remove code duplicates use pytest.parametrize

drapail:
The only duplication I saw was in `valid_url`, so I decided to move it to variable declaration block, instead of using pytest.parametrize (it would be an overkill I think).

app/test_main.py
```diff
@@ -1 +1,39 @@
...
+        mock_has_internet_connection: MagicMock) -> None:
+    mock_has_internet_connection.return_value = False
+    assert can_access_google_page(valid_url) == "Not accessible"
+
+
+@mock.patch("app.main.valid_google_url")
+def test_can_access_google_page_invalid_url_not_ok(
+        mock_valid_google_url: MagicMock) -> None:
+    mock_valid_google_url.return_value = False
+    assert can_access_google_page(invalid_url) == "Not accessible"
```
ArsenPidhoretskyi, line 39: Also use pytest.parametrize
And check all 4 variants:
true - true
true - false
false - true
false - false

ArsenPidhoretskyi (changes requested):

app/test_main.py
drapail: using mark.parametrize together with patch is a bomb, thanks!

ArsenPidhoretskyi (approved):
Well done!

PR #273, opened 2023-02-22
Solution DonchykD

DmytryTong:


app/test_main.py
```diff
@@ -1 +1,53 @@
...
+
+
+@pytest.fixture()
+def mocked_has_connection() -> None:
+    with mock.patch("app.main.has_internet_connection") as mocked_func:
+        yield mocked_func
+
+
+def test_valid_connection_and_url_exist(mocked_valid_google_url: any,
+                                        mocked_has_connection: any
```
amber-marichi, line 21: Your mocked_valid_google_url and mocked_has_connection fixtures both should yield mocks, right? so why dont you type hint it in your tests like "mock" instead of "any"?

app/test_main.py
```diff
@@ -1 +1,53 @@
...
+                                        ) -> None:
+    mocked_valid_google_url.return_value = True
+    mocked_has_connection.return_value = True
+    url = "http//:www.google.com/"
```
amber-marichi, line 25: Are you sure this "url" variables are necessary? They are never used outside function call, so maybe it can be simply removed?

app/test_main.py
```diff
@@ -1 +1,53 @@
...
+    assert can_access_google_page(url) == "Not accessible"
+
+
+def test_invalid_connection_and_invalid_url_exist(mocked_valid_google_url: any,
+                                                  mocked_has_connection: any
+                                                  ) -> None:
+    mocked_valid_google_url.return_value = False
+    mocked_has_connection.return_value = False
+    url = "http//:www.google.com/"
+    assert can_access_google_page(url) == "Not accessible"
```
amber-marichi, line 53: I'm not sure this test is needed. Since "and" is a short-circuit valid_google_url func will never be called if has_internet_connection returns False. You could just replace this whole test with "mocked_valid_google_url.assert_not_called()" in the test with no connection imo.

amber-marichi (changes requested):

app/test_main.py
DmytryTong: Ok)

app/test_main.py
DmytryTong: Yes, "url" is better be there, cause if someone would be needed to change url (in different tests, for example), they can change "url" variable. And this makes code easier to read) And there was a problem in similar case with some mentors (I should add a variable in that case, so I added it here too), so it's up to them: someone likes variables there, someone - don't)

app/test_main.py
DmytryTong: Yep, that is good one: fixed) Thanks)

amber-marichi: what about other test functions? im afraid i cant see it was changed anywhere beside this test :) just write me back here in case you've already done it and im missing something.

app/test_main.py
amber-marichi: i see. allright then :)

app/test_main.py
```diff
@@ -1 +1,53 @@
...
+
+@pytest.fixture()
+def mocked_has_connection() -> None:
+    with mock.patch("app.main.has_internet_connection") as mocked_func:
+        yield mocked_func
+
+
+def test_valid_connection_and_url_exist(mocked_valid_google_url: any,
+                                        mocked_has_connection: any
+                                        ) -> None:
```
Serhii-Leonenko, line 22: Fix indentation, the same in all other similar cases

app/test_main.py
Serhii-Leonenko: In this case, you can make your variable a constant, because you use the same value everywhere and it doesn't matter for your tests what url you use

app/test_main.py
```diff
@@ -1 +1,53 @@
-# write your code here
+import pytest
+from unittest import mock
+
+
```
Serhii-Leonenko, line 4: one blank line is enough

app/test_main.py
Serhii-Leonenko: it is still not fixed in all other similar cases

Serhii-Leonenko (changes requested):

app/test_main.py
DmytryTong: Ok - done.

app/test_main.py
DmytryTong: Fixed)

app/test_main.py
DmytryTong: Fixed)

app/test_main.py
DmytryTong: Fixed)

app/test_main.py
```diff
@@ -1 +1,58 @@
...
+@pytest.fixture()
+def mocked_has_connection() -> None:
+    with mock.patch("app.main.has_internet_connection") as mocked_func:
+        yield mocked_func
+
+
+def test_valid_connection_and_url_exist(
+        mocked_valid_google_url: mock,
+        mocked_has_connection: mock) -> None:
```
Serhii-Leonenko, line 23: You should move the last bracket to the new line as well )


```suggestion
def test_valid_connection_and_url_exist(
        mocked_valid_google_url: mock,
        mocked_has_connection: mock
) -> None:
```

app/test_main.py
```diff
@@ -1 +1,58 @@
-# write your code here
+import pytest
+from unittest import mock
```
Serhii-Leonenko, line 2: Imports should be grouped in the following order:

Standard library imports (unittest.mock module)
Related third-party imports (pytest module)
Local application/library specific imports.

You should put a blank line between each group of imports.

Serhii-Leonenko (changes requested):

app/test_main.py
DmytryTong: Fixed)

app/test_main.py
DmytryTong: Fixed)

amber-marichi (approved):
apr

YuriiKindrat (approved):

PR #276, opened 2023-02-22
Solution

kuznetsb:


RuslanLopatin (approved):

app/test_main.py
```diff
@@ -1 +1,40 @@
...
+
+@pytest.fixture()
+def mocked_internet_connection() -> None:
+    with mock.patch("app.main.has_internet_connection") as mock_connection:
+        yield mock_connection
+
+
+def test_should_access_google_page(
+        mocked_url_validation: object,
+        mocked_internet_connection: object
```
Serhii-Leonenko, line 22: it's better to use more specific annotation

app/test_main.py
```diff
@@ -1 +1,40 @@
...
+) -> None:
+    mocked_url_validation.return_value = False
+    assert can_access_google_page("maate.academy") == "Not accessible"
+
+
+def test_should_not_access_if_no_connection(
+        mocked_internet_connection: object
+) -> None:
+    mocked_internet_connection.return_value = False
+    assert can_access_google_page("facebook.com") == "Not accessible"
```
Serhii-Leonenko, line 40: You should use both mocks in all tests and check  when both are False also

Serhii-Leonenko (changes requested):

YuriiKindrat (changes requested):

PR #283, opened 2023-02-23
'Solution'

oksanaaam:


ProdyRodion (approved):
great job!

app/test_main.py
```diff
@@ -1 +1,46 @@
-# write your code here
+from unittest import mock
+from app.main import can_access_google_page
```
YuriiKindrat, line 2: add blank line between this 2 imports

app/test_main.py
```diff
@@ -1 +1,46 @@
...
+
+@mock.patch("app.main.has_internet_connection", return_value=True)
+@mock.patch("app.main.valid_google_url", return_value=False)
+def test_invalid_url_and_has_internet_connection(
+        mock_valid_url: mock,
+        mock_has_connection: mock
+) -> None:
+    assert can_access_google_page(
+        "https://www.youtube.com/"
+    ) == "Not accessible"
```
YuriiKindrat, line 46: no need to use different urls for tests, u can make 1 constant with url and use it every where

YuriiKindrat (changes requested):

app/test_main.py
Serhii-Leonenko: still not fixed 

image

Serhii-Leonenko (changes requested):

YuriiKindrat (approved):

PR #285, opened 2023-02-23
Solution

ochernyi:


app/test_main.py
```diff
@@ -1 +1,27 @@
...
+@mock.patch("app.main.valid_google_url")
+@mock.patch("app.main.has_internet_connection")
+def test_valid_url_when_connection_exists(
+        valid_google_url: bool,
+        has_internet_connection: bool
+) -> None:
+    valid_google_url.return_value = False
+    has_internet_connection.return_value = True
+
+    assert can_access_google_page("google.com") == "Not accessible"
```
ArsenPidhoretskyi, line 27: Use pytest.parametrize,
because the code almost the same and check all 4 cases
true - true
true - false
false - true
false - false
Fow not function is not covereged

ArsenPidhoretskyi (changes requested):

app/test_main.py
```diff
@@ -1 +1,53 @@
...
+#     return mocked_google_url
+
+
+# @pytest.fixture
+# @mock.patch("app.main.has_internet_connection")
+# def mocked_internet_connection(
+#         mocked_internet_connection,
+# ) -> None:
+#     return mocked_internet_connection
+#
```
Grayder0152, line 22: pls remove comments

Grayder0152 (approved):

PR #286, opened 2023-02-23
Solution

oigod:


Danil228 (approved):

app/test_main.py
```diff
@@ -1 +1,54 @@
-# write your code here
+import unittest
+from unittest import mock
+
+import app.main
+
+
+class TestCanAccessGooglePage(unittest.TestCase):
```
YuriiKindrat, line 7: no need to use classes here

app/test_main.py
```diff
@@ -1 +1,54 @@
...
+    @mock.patch("app.main.has_internet_connection")
+    @mock.patch("app.main.valid_google_url")
+    def test_accessible(
+            self,
+            mock_valid_url: bool,
+            mock_has_internet: bool
+    ) -> None:
+        mock_valid_url.return_value = True
+        mock_has_internet.return_value = True
+        result = app.main.can_access_google_page("https://www.google.com")
```
YuriiKindrat, line 17: image
also no need to change url in ur tests, u can write it as constant and use everywhere

YuriiKindrat (changes requested):

app/test_main.py
```diff
@@ -1 +1,48 @@
-# write your code here
+from unittest import mock
+from typing import Any
+
+import app.main
+
+google_url = "https://www.google.com"
+invalid_url = "https://www.invalidurl.com"
```
YuriiKindrat, line 7: there is no need to use different urls in this tests

YuriiKindrat (approved):

PR #288, opened 2023-02-23
Solution

TeliksPG:


viktoria-rybenchuk (approved):

app/test_main.py
```diff
@@ -1 +1,27 @@
...
+@mock.patch("app.main.valid_google_url", return_value=True)
+@mock.patch("app.main.has_internet_connection", return_value=False)
+def test_with_no_internet_connection(url: mock, internet: mock) -> None:
+    assert can_access_google_page("https://www.google.com") == "Not accessible"
+
+
+@mock.patch("app.main.valid_google_url", return_value=False)
+@mock.patch("app.main.has_internet_connection", return_value=False)
+def test_with_no_connection_and_wrong_url(url: mock, internet: mock) -> None:
+    assert can_access_google_page("https:/www.smorfle.com") == "Not accessible"
```
YuriiKindrat, line 27: no need to use different urls for tests, u can add a constant with url and use it everywhere

YuriiKindrat (changes requested):

app/test_main.py
ProdyRodion: but thats just for fun, thay may be empty

Serhii-Leonenko (approved):

PR #290, opened 2023-02-23
'Solution'

alexlega:


ostapT (approved):

YuriiKindrat (changes requested):
no need to create url in every test, u can create constant and use it everywhere

app/test_main.py
```diff
@@ -1 +1,33 @@
...
+
+@mock.patch("app.main.valid_google_url")
+@mock.patch("app.main.has_internet_connection")
+@pytest.mark.parametrize(
+    "validate_google_url, has_internet_connection, expected_result",
+    [
+        (True, True, "Accessible"),
+        (True, False, "Not accessible"),
+        (False, True, "Not accessible"),
+        (False, False, "Not accessible"),
```
wQuelS, line 14: Add an error message. It's good when the user understands what the problem is. Fix everywhere

wQuelS (changes requested):

YuriiKindrat (approved):

PR #292, opened 2023-02-23
Solution

oksana0410:


inrock1 (approved):

app/test_main.py
```diff
@@ -1 +1,42 @@
-# write your code here
+from unittest import TestCase
+from unittest.mock import patch
+from app.main import can_access_google_page
+from typing import Callable
```
YuriiKindrat, line 4: fix imports 
https://peps.python.org/pep-0008/#:~:text=Imports%20are%20always,library%20specific%20imports.

app/test_main.py
```diff
@@ -1 +1,42 @@
...
+
+
+class TestAccess(TestCase):
```
YuriiKindrat, line 7: no need to use classes here

app/test_main.py
```diff
@@ -1 +1,42 @@
...
+    @patch("app.main.has_internet_connection")
+    def test_is_not_accessible_because_of_url(
+            self,
+            mock_internet: Callable,
+            mock_url: Callable
+    ) -> None:
+        mock_internet.return_value = True
+        mock_url.return_value = False
+        assert (can_access_google_page("http://www.notgoogle.com")
+                == "Not accessible")
```
YuriiKindrat, line 42: no need to use different urls for ur tests, u can make constant from url and use it ur tests

YuriiKindrat (changes requested):
What about False/False condition ?

app/test_main.py
```diff
@@ -1 +1,36 @@
...
+from typing import Callable
+from unittest.mock import patch
+from app.main import can_access_google_page
+
+
+url = "http://www.google.com"
+
+
+@patch("app.main.valid_google_url")
+@patch("app.main.has_internet_connection")
```
wQuelS, line 10: Imagine having a fixture and not writing those decorators again and again..

app/test_main.py
```diff
@@ -1 +1,36 @@
...
+def test_is_accessible(mock_internet: Callable, mock_url: Callable) -> None:
+    mock_internet.return_value = True
+    mock_url.return_value = True
+    assert can_access_google_page(url) == "Accessible"
```
wQuelS, line 14: Add an error message. It's good when the user understands what the problem is. Fix everywhere

wQuelS (changes requested):

app/test_main.py
YuriiKindrat: what about this test?

app/test_main.py
```diff
@@ -1 +1,59 @@
...
+        "Fix your internet connection!"
+
+
+def test_is_not_accessible_because_of_url_and_internet_conection(
+        mock_valid_url: Callable,
+        mock_internet_available: Callable
+) -> None:
+    mock_valid_url.return_value = False
+    mock_internet_available.return_value = False
+    assert can_access_google_page(url) == "Not accessible",\
```
YuriiKindrat, line 58: try to avoid using " \ " add brackets instead

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #294, opened 2023-02-23
'Solution'

AnnaBoichenko:


Zenionoi (approved):
In my opinion, there is a lot of code duplication, you can use `@pytest.mark.parametrize` or `@pytest.fixture()` but data definition in the test seems to be correct as well, so I leave it as just a comment with approve

app/test_main.py
```diff
@@ -1 +1,52 @@
-# write your code here
+from unittest import mock
+
+from app.main import can_access_google_page
+
+
+def test_have_internet_and_correct_url() -> None:
+    with (
+        mock.patch("app.main.has_internet_connection") as valid_url_mock,
+        mock.patch("app.main.valid_google_url") as has_internet_mock,
```
wQuelS, line 9: As it was said, you have broken DRY principle. There are a lot of code duplication. Remember, that you can use fixtures

app/test_main.py
```diff
@@ -1 +1,52 @@
...
+    ):
+        has_internet_mock.return_value = True
+        valid_url_mock.return_value = True
+        assert can_access_google_page("https://www.google.com") == "Accessible"
```
wQuelS, line 13: Add an error message. It's good when the user understands what the problem is. Fix everywhere

wQuelS (changes requested):

PR #295, opened 2023-02-23
Solution

VladyslavCherkashyn:


usrofgh (approved):

app/test_main.py
```diff
@@ -1 +1,45 @@
-# write your code here
+from unittest import mock
+from app.main import can_access_google_page
+
+
+@mock.patch("app.main.has_internet_connection")
+@mock.patch("app.main.valid_google_url")
```
wQuelS, line 6: Imagine having a `fixture` and not writing those decorators again and again..

app/test_main.py
```diff
@@ -1 +1,45 @@
...
+def test_if_both_true(
+        mock_valid_google_url: mock,
+        mock_has_internet_connection: mock
+) -> None:
+    mock_valid_google_url.return_value = True
+    mock_has_internet_connection.return_value = True
+    assert can_access_google_page("lol") == "Accessible"
```
wQuelS, line 13: Add an error message. It's good when the user understands what the problem is. Fix everywhere

wQuelS (changes requested):

app/test_main.py
VladyslavCherkashyn: I apologize, I did not pay attention when committed, shame on me.

YuriiKindrat (changes requested):
Add an error message. It's good when the user understands what the problem is

wQuelS (approved):
GJ

PR #296, opened 2023-02-23
Solution

viktoria-rybenchuk:


ykopatko (approved):
GJ. However, you may  use @pytest.mark.parametrize in order optimaze your code

Serhii-Leonenko (changes requested):
Agree with the previous comment, do according to it

YuriiKindrat (approved):

PR #297, opened 2023-02-23
Solution

NazarHladaniuk:


Anatolii-Poznyak (approved):

app/test_main.py
```diff
@@ -1 +1,31 @@
...
+import unittest
+from unittest.mock import patch
+from datetime import datetime
+from app.main import can_access_google_page
+
+
+class TestCanAccessPage(unittest.TestCase):
+    def test_valid_url_and_internet_connection(self) -> None:
+        with patch("app.main.valid_google_url", return_value=True):
+            with patch("app.main.has_internet_connection", return_value=True):
```
wQuelS, line 10: if you are using unitests, consider using `SetUp` not to write those patches again and again. With `pytest` you could simply use `fixture`

app/test_main.py
```diff
@@ -1 +1,31 @@
...
+                url = "https://www.google.com/"
+                result = can_access_google_page(url)
+                self.assertEqual(result, "Accessible")
+
+    def test_invalid_url_valid_internet_connection(self) -> None:
+        with patch("app.main.valid_google_url", return_value=False):
+            with patch("app.main.has_internet_connection", return_value=True):
+                url = "https://www.Geag0Le.com/"
+                result = can_access_google_page(url)
+                self.assertEqual(result, "Not accessible")
```
wQuelS, line 20: Add an error message. It's good when the user understands what the problem is. Fix everywhere

wQuelS (changes requested):

app/test_main.py
```diff
@@ -1 +1,39 @@
-# write your code here
+import unittest
+from unittest.mock import patch
+from datetime import datetime
+from app.main import can_access_google_page
```
Serhii-Leonenko, line 4: You should split different groups of imports with a blank line

app/test_main.py
```diff
@@ -1 +1,39 @@
...
+        ):
+            result = can_access_google_page(self.invalid_url)
+            self.assertEqual(result, "Not accessible",
+                             f"Expected 'Not accessible' but got '{result}'.")
+
+    def test_out_of_hours(self) -> None:
+        with (
+            patch("app.main.valid_google_url", return_value=True),
+            patch("app.main.has_internet_connection", return_value=False),
+            patch("app.main.datetime.datetime", return_value=self.current_time)
```
Serhii-Leonenko, line 35: Do you really need this patch?

Serhii-Leonenko (changes requested):
Add test for the case when invalid url and no connection

YuriiKindrat (changes requested):

wQuelS (approved):
GJ

PR #298, opened 2023-02-23
solution

AlessaQ:


destr34 (approved):

app/test_main.py
```diff
@@ -1 +1,55 @@
...
+        yield mock_time
+
+
+def test_access_google_page_true(
+    mocked_valid_google_url: Callable,
+    mocked_has_internet_connection: Callable
+) -> None:
+    mocked_valid_google_url.return_value = True
+    mocked_has_internet_connection.return_value = True
+    assert can_access_google_page("www.google.com") == "Accessible"
```
wQuelS, line 28: Add an error message. It's good when the user understands what the problem is. Fix everywhere

wQuelS (changes requested):

app/test_main.py
AlessaQ: Added

Serhii-Leonenko: You should add it here as well. And use more descriptive messages, for example, a web page should be Accessible if ....

app/test_main.py
```diff
@@ -1 +1,57 @@
-# write your code here
+import pytest
+
+from app.main import can_access_google_page
+
+from unittest import mock
+
+from typing import Callable
```
Serhii-Leonenko, line 7: image

Serhii-Leonenko (changes requested):

app/test_main.py
AlessaQ: Messages changed

app/test_main.py
AlessaQ: fixed

app/test_main.py
```diff
@@ -1 +1,56 @@
...
+    with mock.patch("app.main.has_internet_connection") as mock_time:
+        yield mock_time
+
+
+def test_access_google_page_true(
+    mocked_valid_google_url: Callable, mocked_has_internet_connection: Callable
+) -> None:
+    mocked_valid_google_url.return_value = True
+    mocked_has_internet_connection.return_value = True
+    assert can_access_google_page("www.google.com") == "Accessible"
```
Serhii-Leonenko, line 26: Add a message for this test  as well

Serhii-Leonenko (changes requested):

app/test_main.py
AlessaQ: Added

PR #299, opened 2023-02-23
'Solution'

inrock1:


me1nyk (approved):

app/test_main.py
```diff
@@ -1 +1,31 @@
...
+
+from app.main import can_access_google_page
+
+
+@pytest.mark.parametrize(
+    "valid_url, connection, access",
+    [
+        pytest.param(True, True, "Accessible",
+                     id="Test can access. All valid"),
+        pytest.param(True, False, "Not accessible",
```
wQuelS, line 12: Add an error message. It's good when the user understands what the problem is. Fix everywhere

wQuelS (changes requested):

app/test_main.py
inrock1: was added more detailed id

app/test_main.py
```diff
@@ -1 +1,39 @@
-# write your code here
+from unittest import mock
+import pytest
```
Serhii-Leonenko, line 2: You should put a blank line between each group of imports.

Serhii-Leonenko (approved):

PR #301, opened 2023-02-23
'Solution'

ykopatko:


NickSavchyn (approved):

app/test_main.py
```diff
@@ -1 +1,29 @@
-# write your code here
+import pytest
+
+from unittest import mock
+
+from .main import can_access_google_page
```
wQuelS, line 5: use full path. I mean `app.main`

app/test_main.py
```diff
@@ -1 +1,29 @@
...
+
+
+@pytest.mark.parametrize(
+    "valid_url,connection,result",
+    [
+        (True, False, "Not accessible"),
```
wQuelS, line 11: Add an error message. It's good when the user understands what the problem is. Fix everywhere

wQuelS (changes requested):

app/test_main.py
ykopatko: > Add an error message. It's good when the user understands what the problem is. Fix everywhere

Do I understand this right, that you mean IDs? I have added them

Serhii-Leonenko (approved):
Well done!

PR #303, opened 2023-02-23
Solution

RomanRega:


app/test_main.py
```diff
@@ -1 +1,38 @@
-# write your code here
+from app.main import can_access_google_page
+
+from unittest import mock
+
+from unittest.mock import MagicMock
```
Serhii-Leonenko, line 5: image

app/test_main.py
```diff
@@ -1 +1,38 @@
...
+
+@mock.patch("app.main.valid_google_url")
+@mock.patch("app.main.has_internet_connection")
+def test_no_internet_valid_url(
+        valid_google_url: MagicMock,
+        has_internet_connection: MagicMock
+) -> None:
+    valid_google_url.return_value = True
+    has_internet_connection.return_value = False
+    assert can_access_google_page("http://www.google.com") == "Not accessible"
```
Serhii-Leonenko, line 38: What about the case when both values are False?

Also, use pytest.mark.parametrize to remove code duplicating and add messages

wQuelS (approved):
Good

PR #304, opened 2023-02-23
solution

usrofgh:


app/test_main.py
```diff
@@ -1 +1,44 @@
-# write your code here
+from unittest import mock
+import pytest as pytest
```
Serhii-Leonenko, line 2: `import pytest as pytest` - interesting, what is the purpose?)

Also, different groups of imports you should split with a blank line

app/test_main.py
```diff
@@ -1 +1,44 @@
...
+def mocked_has_internet_connection() -> mock.Mock:
+    with mock.patch("app.main.has_internet_connection") \
+            as mock_has_internet_connection:
+        yield mock_has_internet_connection
+
+
+@pytest.fixture()
+def mocked_valid_google_url() -> mock.Mock:
+    with mock.patch("app.main.valid_google_url") as mock_valid_google_url:
+        yield mock_valid_google_url
```
Serhii-Leonenko, line 16: Try to unite into one function

Serhii-Leonenko (changes requested):

wQuelS (approved):
Good

PR #307, opened 2023-02-23
Solution

ZhAlexR:


app/test_main.py
```diff
@@ -1 +1,35 @@
-# write your code here
+import pytest
+from unittest import mock
```
Serhii-Leonenko, line 2: image

app/test_main.py
```diff
@@ -1 +1,35 @@
...
+
+from app.main import can_access_google_page
+
+use_case_data = [
+    (True, True, "Accessible"),
+    (True, False, "Not accessible"),
+    (False, True, "Not accessible"),
```
Serhii-Leonenko, line 9: What about the case when both values are False?

PR #311, opened 2023-02-23
Solution

me1nyk:


app/test_main.py
```diff
@@ -1 +1,35 @@
...
+
+@mock.patch("app.main.valid_google_url", return_value=True)
+@mock.patch("app.main.has_internet_connection", return_value=False)
+def test_can_access_google_page_not_accessible(
+        google_url: mock,
+        internet_connection: mock
+) -> None:
+    assert can_access_google_page(
+        "https://www.google.com"
+    ) == "Not accessible"
```
Serhii-Leonenko, line 35: Try to use patest.mark.parametrize to remove duplicating, also add a check for case when both values are False and add messages of course

Serhii-Leonenko (changes requested):

wQuelS (approved):
Nice

PR #322, opened 2023-03-02
'Solution'

Mykhailo-Benzel:


ArsenPidhoretskyi (changes requested):
Use pytest.parametrize try to avoid duplicate of code
Also cover by test when both are False

ArsenPidhoretskyi (approved):
Well done!

PR #324, opened 2023-03-03
'Solution'

natalie-goriela:


ArsenPidhoretskyi (changes requested):
Function is not covered for return values

> This function takes url of the site, returns "Accessible" if url is valid to access the Google home page and it has internet connection, else it returns "Not accessible".

olena-hul (approved):

PR #325, opened 2023-03-04
Solution

StreetNik:


app/test_main.py
```diff
@@ -1 +1,58 @@
...
+
+
+def test_has_internet_connection_was_called(mocked_internet: Callable) -> None:
+    can_access_google_page("http://youtube.com/")
+    mocked_internet.assert_called_once_with()
+
+
+def test_valid_google_url_was_called(mocked_url: Callable) -> None:
+    can_access_google_page("google.com/")
+    mocked_url.assert_called_once_with("google.com/")
```
ArsenPidhoretskyi, line 58: You can remove this tests

ArsenPidhoretskyi: It failed because due to code:
`if has_internet_connection() and valid_google_url(url):`
has_internet_connection() will return False, and checking for valid_google_url will not enter

ArsenPidhoretskyi (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #348, opened 2023-03-23
Solution

WalDragnir:


vasylhnatiuk (changes requested):
it would be good to add test when `mocked_valid_google_url.return_value` is False and `has_internet_connection.return_value` is False

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #351, opened 2023-04-03
Solution

anastasia-martyniuk:


app/test_main.py
```diff
@@ -1 +1,64 @@
-# write your code here
+from unittest import mock
+
+import pytest
+
+from app.main import can_access_google_page
+
+
+@pytest.fixture()
+def mocked_valid_google_url() -> None:
```
EdAlekseiev, line 9: `None` is incorrect type hint, because this function returns `MagicMock` objects

app/test_main.py
```diff
@@ -1 +1,64 @@
...
+def test_should_return_not_accessible_when_url_is_not_valid_but_has_connection(
+    mocked_valid_google_url: mock, mocked_has_internet_connection: mock
+) -> None:
+    mocked_valid_google_url.return_value = False
+    mocked_has_internet_connection.return_value = True
+
+    assert (
+        can_access_google_page(url="https://www.google.com/")
+        == "Not accessible"
+    )
```
EdAlekseiev, line 64: Please, use `pytest.mark.parametrize` instead of code duplicate.

ArsenPidhoretskyi (approved):
Well done!

PR #357, opened 2023-04-10
Solution

Anyta17:


app/test_main.py
```diff
@@ -1 +1,17 @@
...
+            assert can_access_google_page("https://www.google.com")\
+                   == "Not accessible"
+
+        with patch("app.main.has_internet_connection", return_value=True):
+            assert can_access_google_page("https://www.google.com")\
+                   == "Accessible"
+
+        with patch("app.main.valid_google_url", return_value=False):
+            assert can_access_google_page("https://invalidurl.com")\
+                   == "Not accessible"
```
ArsenPidhoretskyi, line 17: Use pytest.parametrize and check 4 different cases
true-true
false-true
true-false
false-false

ArsenPidhoretskyi (changes requested):

PR #365, opened 2023-04-19
or better create separated test for every assert?

DmytroShvetsDev:


app/test_main.py
```diff
@@ -1 +1,23 @@
...
+    mock_has_internet_connection.return_value = True
+    assert can_access_google_page("https://www.google.com") == "Not accessible"
+
+    mock_valid_google_url.return_value = True
+    mock_has_internet_connection.return_value = False
+    assert can_access_google_page("https://www.google.com") == "Not accessible"
+
+    mock_valid_google_url.return_value = False
+    mock_has_internet_connection.return_value = False
+    assert can_access_google_page("https://www.google.com") == "Not accessible"
```
ArsenPidhoretskyi, line 23: Use @pytest.parametrize No need to duplicate code

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #374, opened 2023-05-09
'Solution'

VTeteruk:


app/test_main.py
```diff
@@ -1 +1,36 @@
...
+        with mock.patch("app.main.has_internet_connection") as connection:
+            with mock.patch("app.main.valid_google_url") as valid_google_url:
+                valid_google_url.return_value = valid_result
+                connection.return_value = connection_result
+
+                assert (
+                    can_access_google_page("https://www.google.com.ua")
+                    == expected_result
+                )
+                if connection.return_value:
```
vasylhnatiuk, line 30: why do you need this? always connection.return_value = connection_result, the below code is redundant

vasylhnatiuk (changes requested):

PR #380, opened 2023-05-12
Solution

Dimdimior:


app/test_main.py
```diff
@@ -1 +1,42 @@
...
+            "Not accessible",
+            id="No access without url_response"
+        ),
+        pytest.param(
+            True,
+            True,
+            "Accessible",
+            id="Access is possible"
+        ),
+    ]
```
ArsenPidhoretskyi, line 28: Add test for False and False

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #387, opened 2023-05-18
Solution

Munkoor:


vasylhnatiuk (changes requested):
Add a test if two mocks are false.

Grayder0152 (approved):

PR #388, opened 2023-05-21
Solution

markus2812:


vasylhnatiuk (changes requested):
image
it will be better to create a few separate tests or use parametrsion

app/test_main.py
```diff
@@ -1 +1,23 @@
...
+        assert result == "Accessible"
+
+        mock_has_internet_connection.return_value = False
+        result = can_access_google_page("https://www.google.com")
+        assert result == "Not accessible"
+
+        mock_valid_google_url.return_value = False
+        mock_has_internet_connection.return_value = True
+        result = can_access_google_page("https://www.invalidurl.com")
+        assert result == "Not accessible"
```
ArsenPidhoretskyi, line 23: Use pytest parametrize instead of duplicate code

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #389, opened 2023-05-21
solution

IvanKorshunovE:


app/test_main.py
```diff
@@ -1 +1,27 @@
...
+
+from app.main import can_access_google_page
+
+
+@pytest.mark.parametrize(
+    "page_is_valid,has_internet,result",
+    [
+        (True, False, "Not accessible"),
+        (False, True, "Not accessible"),
+        (True, True, "Accessible")
```
Vasyl-Smutok, line 12: what about `False and False ` ?

Vasyl-Smutok (changes requested):

Vasyl-Smutok (approved):
good Job 👍

PR #395, opened 2023-05-23
py-can-access-google-page-v1

muriakula:


EdAlekseiev (changes requested):
Use @pytest.mark.parametrize. Read more: https://docs.pytest.org/en/7.1.x/how-to/parametrize.html

Bohdan-Salamakha (approved):

PR #401, opened 2023-06-03
Solution

gleblitvinenko:


app/test_main.py
```diff
@@ -1 +1,44 @@
...
+    mocked_valid_google_url.return_value = False
+    assert can_access_google_page("google.com") == "Not accessible"
+
+
+def test_accessible(
+        mocked_valid_google_url: Callable,
+        mocked_has_internet_connection: Callable) -> None:
+    mocked_has_internet_connection.return_value = True
+    mocked_valid_google_url.return_value = True
+    assert can_access_google_page("google.com") == "Accessible"
```
ArsenPidhoretskyi, line 44: Use pytest parametrize. Also 4 cases possible:
True - True
True - False
False - True
False - False

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):
Well done

PR #406, opened 2023-06-12
'Solution'

Artistgatehub:


app/test_main.py
```diff
@@ -1 +1,23 @@
...
+        current_time.return_value = True
+        assert can_access_google_page("") == "Not accessible"
+
+        response.return_value = False
+        current_time.return_value = False
+        assert can_access_google_page("") == "Not accessible"
+
+        response.return_value = True
+        current_time.return_value = False
+        assert can_access_google_page("") == "Not accessible"
```
EdAlekseiev, line 23: Use pytest.mark.parametrize

app/test_main.py
```diff
@@ -1 +1,23 @@
-# write your code here
+from app.main import can_access_google_page
+
+from unittest.mock import patch
+
+
+def test_can_access_google_page_correct_result() -> None:
+    with (patch("app.main.valid_google_url") as response,
+          patch("app.main.has_internet_connection") as current_time):
```
EdAlekseiev, line 8: Use `patch` as decorator instead of context manager

EdAlekseiev (changes requested):

YukihiroSM (approved):

YukihiroSM (changes requested):
Use patch decorator instead of context manager.

PR #407, opened 2023-06-14
Solution

YevheniiIovchu:


AntonShpakovych (approved):
!!! change sequence of decorators - it will be more logical if you first check whether the Internet is available!!!

But you can simplify this using parametrize where u put value for mocks, also for url, and expected_result

app/test_main.py
```diff
@@ -1 +1,30 @@
...
+    mocked_valid.return_value = True
+    assert (
+        can_access_google_page("https://www.youtube.com/") == "Not accessible"
+    )
+
+    mocked_connection.return_value = True
+    mocked_valid.return_value = False
+    assert (
+        can_access_google_page("https:/ww.you1t1u1be.com/") == "Not accessible"
+    )
```
YuriiKindrat, line 30: could u use parametrize here?)
also what about False, Flase case ?

YuriiKindrat (changes requested):

app/test_main.py
```diff
@@ -1 +1,49 @@
...
+        pytest.param(
+            False, True, "https://www.youtube.com/", "Not accessible"
+        ),
+        pytest.param(
+            True, False, "https:/ww.you1t1u1be.com/", "Not accessible"
+        ),
+        pytest.param(
+            False, False, "https:/ww.you1t1u1be.com/", "Not accessible"
+        ),
+    ],
```
YuriiKindrat, line 25: great, but could u make a constant for URL, so u could avoid code repeating ?)

YuriiKindrat (changes requested):

app/test_main.py
```diff
@@ -1 +1,53 @@
...
+
+
+from unittest import mock
+
+
+from app.main import can_access_google_page
+
+
+VALID_URL = "https://www.youtube.com/"
+INVALID_URL = "https:/ww.you1t1u1be.com/"
```
YuriiKindrat, line 11: there is no need to use different urls, but it is up to u )

YuriiKindrat (approved):

PR #409, opened 2023-06-14
'Solution'

ant-komarov:


app/test_main.py
```diff
@@ -1 +1,41 @@
...
+@mock.patch("app.main.valid_google_url")
+@mock.patch("app.main.has_internet_connection")
+def test_can_not_access_with_invalid_url(
+        mocked_connection: Callable,
+        mocked_valid_url: Callable
+) -> None:
+    mocked_connection.return_value = True
+    mocked_valid_url.return_value = False
+
+    assert can_access_google_page("url") == "Not accessible"
```
DmytroTsehelnyk, line 41: Could you make your code DRY here? You can make just 1 test with the same mocks using parametrize.

DmytroTsehelnyk (changes requested):

app/test_main.py
```diff
@@ -1 +1,31 @@
...
+
+@pytest.mark.parametrize(
+    "is_connected,is_valid_url,expected",
+    [
+        (True, True, "Accessible"),
+        (False, True, "Not accessible"),
+        (True, False, "Not accessible"),
+        (False, False, "Not accessible")
+    ]
+)
```
DmytroTsehelnyk, line 18: Add ids with appropriate messages that the programmer will see if test failed

app/test_main.py
```diff
@@ -1 +1,31 @@
...
+@mock.patch("app.main.valid_google_url")
+@mock.patch("app.main.has_internet_connection")
```
DmytroTsehelnyk, line 20: Change the decorators' sequence. 
It will be more logical if you first check whether the user has an internet connection

DmytroTsehelnyk (changes requested):

app/test_main.py
```diff
@@ -1 +1,37 @@
...
+@pytest.mark.parametrize(
+    "is_connected,is_valid_url,expected",
+    [
+        (True, True, "Accessible"),
+        (False, True, "Not accessible"),
+        (True, False, "Not accessible"),
+        (False, False, "Not accessible")
+    ],
+    ids=[
+        "You can access with connection nd valid url",
```
DmytroTsehelnyk, line 19: GJ! Just fix this type here

Oleg-s-s (approved):
:+1:

PR #410, opened 2023-06-14
Solution

AntonShpakovych:


Kirontiko (approved):
I think you can use constant variable to define url and do not repeat them in parameters, but I think it's up to you

AntonShpakovych:
> I think you can use constant variable to define url and do not repeat them in parameters, but I think it's up to you

Good idea, but I don't know where I can store this constants. Of course I can create some file instead of store constant  in `test_main.py` or even more create `valid/invalid` fixtures.  Maybe mentors can help with choosing the best option)

app/test_main.py
```diff
@@ -1 +1,57 @@
...
+
+import pytest
+
+from app.main import can_access_google_page
+
+
+class TestCanAccessGooglePage:
+    @pytest.mark.parametrize(
+        "connection_mock,url_mock,expected_result,url",
+        [
```
Oleg-s-s, line 12:
```suggestion
    TEST_URL = ...

    @pytest.mark.parametrize(
        "connection_mock,url_mock,expected_result,url",
        [
```

Oleg-s-s (changes requested):

PR #411, opened 2023-06-14
Solution

diana-shyrokikh:


YKonoplyov (approved):
Good job, I`m impressed

Oleg-s-s (changes requested):
You can combine these test and add id's, for url you can create constant to not repeat yourself

YuriiKindrat (approved):

PR #413, opened 2023-06-15
Solution

Kirontiko:


yakhlamova (approved):

app/test_main.py
```diff
@@ -1 +1,54 @@
...
+            "https://gitmug.com/3140,Kirontiko/",
+            "Not accessible",
+            id="not accessible when url is not valid"
+        ),
+        pytest.param(
+            False,
+            False,
+            "https://gitmug.com/314-K,l;.irontiko/",
+            "Not accessible",
+            id="not accessible when url is not valid and no connection"
```
YuriiKindrat, line 38: why do u need to use different urls here? 
could u use only one url and pass it directly to ur func ?

YuriiKindrat (changes requested):

app/test_main.py
Kirontiko: ok, will fix right now:)

app/test_main.py
```diff
@@ -1 +1,57 @@
-# write your code here
+from unittest import mock
+from typing import Callable
+
+import pytest
+
+from app.main import can_access_google_page
+
+
+TEST_URL = "https://github.com/Kirontiko/"
```
YuriiKindrat, line 9: could u pass it directly to the func ? on line 57)

YuriiKindrat (changes requested):

app/test_main.py
Kirontiko: Oh, sorry, I understood wrong

YuriiKindrat (approved):

PR #414, opened 2023-06-15
solution

oleksiikolii:


app/test_main.py
```diff
@@ -11,10 +11,10 @@
     ("https://www.google.com", False, True, "Not accessible")
 ])
 def test_valid_url_and_connection_exists(monkeypatch: MonkeyPatch,
```
Altahoma, line 13: Perhaps this indentation style will provide better readability for function with long names:
```
def test_valid_url_and_connection_exists(
    monkeypatch: MonkeyPatch,
    url: str,
    expected: str,
    valid: bool,
    connection: bool
) -> None:
    pass
```

app/test_main.py
```diff
@@ -11,10 +11,10 @@
...
-                           url: str,
-                           expected: str,
-                           valid: bool,
-                           connection: bool) -> None:
+                                         url: str,
+                                         expected: str,
+                                         valid: bool,
+                                         connection: bool) -> None:
     with monkeypatch.context() as m:
```
Altahoma, line 18: I'm sure you could choose a better name for the variable than just `m`.

Altahoma (changes requested):

app/test_main.py
Altahoma: Ooo, I see that this is a template from the documentation 🙃

Altahoma (approved):

app/test_main.py
Oleg-s-s: Agree, on the other hand you should keep the same bracket open/close style f.e.:
(something,
something)
or
(
something,
something
)

app/test_main.py
```diff
@@ -1 +1,24 @@
...
+
+
+from app.main import can_access_google_page
+
+
+@pytest.mark.parametrize("url,valid,connection,expected", [
+    ("https://www.google.com", True, True, "Accessible"),
+    ("https://www.google.com", True, False, "Not accessible"),
+    ("https://www.google.com", False, True, "Not accessible"),
+    ("https://www.google.com", False, False, "Not accessible")
```
Oleg-s-s, line 12: Use a constant for url to avoid code repeating

app/test_main.py
```diff
@@ -1 +1,26 @@
...
+from pytest import MonkeyPatch
+
+
+from app.main import can_access_google_page
+
+
+URL = "https://www.google.com"
+
+
+@pytest.mark.parametrize("url,valid,connection,expected", [
```
LiudmylaKulinchenko, line 11: Why do you need to pass `url` in the function? You don't use it there

LiudmylaKulinchenko (changes requested):

YuriiKindrat (approved):

PR #415, opened 2023-06-15
Solution

YKonoplyov:


app/test_main.py
```diff
@@ -1 +1,41 @@
-# write your code here
+import pytest
+from unittest import mock
```
oleksiikolii, line 2: it is not recommended to use different frameworks in one test

oleksiikolii (changes requested):

oleksiikolii (approved):
Good job

app/test_main.py
```diff
@@ -1 +1,36 @@
...
+
+@mark.parametrize(
+    "url,url_validation,internet,result",
+    [
+        ("Not google url", False, True, "Not accessible"),
+        ("Some google url", True, True, "Accessible"),
+        ("Not google url", False, True, "Not accessible"),
+        ("Some google url", True, True, "Accessible"),
+        ("Not google url", False, False, "Not accessible"),
+        ("Some google url", True, False, "Not accessible"),
```
Oleg-s-s, line 15: Try to use constant for url to improve readability

Oleg-s-s (changes requested):

app/test_main.py
```diff
@@ -1 +1,39 @@
...
+@mark.parametrize(
+    "url, url_validation,internet,result",
+    [
+        (GOOGLE_URL, False, True, "Not accessible"),
+        (GOOGLE_URL, True, True, "Accessible"),
+        (NOT_GOOGLE_URL, False, True, "Not accessible"),
+        (GOOGLE_URL, True, True, "Accessible"),
+        (NOT_GOOGLE_URL, False, False, "Not accessible"),
+        (GOOGLE_URL, True, False, "Not accessible"),
+    ],
```
YuriiKindrat, line 19: why do u need different "urls" for monkeypatch ?
it seems to me, u can make here only 4 tests for all cases

YuriiKindrat (changes requested):

YuriiKindrat (changes requested):

app/test_main.py
```diff
@@ -1 +1,31 @@
...
+def test_can_access_google_page(url_validation: bool,
+                                internet: bool,
+                                result: str,
+                                monkeypatch: MonkeyPatch) -> None:
+    url = "some_url"
+    monkeypatch.setattr("app.main.valid_google_url",
+                        lambda *args: url_validation)
+    monkeypatch.setattr("app.main.has_internet_connection",
+                        lambda: internet)
+    assert can_access_google_page(url) == result
```
YuriiKindrat, line 31: it seems to me there is no need to create additional variable for the url, u can just pass string directly to the func

YuriiKindrat (approved):

PR #416, opened 2023-06-15
'Solution'

yakhlamova:


AntonShpakovych (approved):

AntonShpakovych (changes requested):
add test for False False

yakhlamova:
> add test for False False

Is it so necessary? If we have no internet connection we have no chance to use validation of url or I misunderstand?

AntonShpakovych:
> > add test for False False
> 
> Is it so necessary? If we have no internet connection we have no chance to use validation of url or I misunderstand?
For us its understandable but we need check function

AntonShpakovych (approved):

Oleg-s-s (changes requested):
Can you make a constant for url to make ur code more readable?

app/test_main.py
```diff
@@ -1 +1,56 @@
...
+            valid_url,
+            "Not accessible",
+            id="not accessible when URL is invalid and no internet connection"
+        ),
+
+    ]
+)
+@mock.patch("app.main.valid_google_url")
+@mock.patch("app.main.has_internet_connection")
+def test_can_access_google_page(mock_internet_connection: Any,
```
LiudmylaKulinchenko, line 48: Use Callable for mocked functions in type annotations

LiudmylaKulinchenko (approved):
Well done!

PR #417, opened 2023-06-15
Solution auto

flatchers:


flatchers:
Oops... thought that will be auto checking

vasylhnatiuk (changes requested):
add tests for             (True, True, "Accessible"),(False, False, "Not accessible")

Grayder0152 (approved):

PR #418, opened 2023-06-15
Implemented tests for can_access_google_page()

Altahoma:


ant-komarov (approved):

app/test_main.py
```diff
@@ -1 +1,33 @@
...
+            id="Access to the page if all requirements are followed"),
+        param(
+            "open google please", False, True, "Not accessible",
+            id="The page cannot be opened if the URL is incorrect"),
+        param(
+            "https://www.google.com/", True, False, "Not accessible",
+            id="The page cannot be accessed if no Internet connection"),
+        param(
+            "ok google, open google", False, False, "Not accessible",
+            id="The page cannot be accessed if all requirements are failed"),
```
YuriiKindrat, line 20: why do u need to use different urls here? could u use only one url and pass it directly to ur func ?

YuriiKindrat (changes requested):

app/test_main.py
YuriiKindrat: not at all, when u are using monkeypatch))
there is separate func to check if valid/invalid url, so there is no need to write valid/invalid urls )

YuriiKindrat (changes requested):

app/test_main.py
Altahoma: It's really interesting, I thought any useful information in the tests = for better, even if it was not significant at first glance.

YuriiKindrat: maybe so, but definitely not violating the DRY principle and occupying unnecessary memory with strings)
also test is checking how fucntion will work with different results from other funcs, so u can use the same url for every test, and it is absolutly not important what url u'll use here

YuriiKindrat (changes requested):

Oleg-s-s (approved):
:+1:

PR #420, opened 2023-06-15
Solution

Andriy-Sydorenko:


imelnyk007 (approved):

app/test_main.py
```diff
@@ -1 +1,46 @@
...
+        ("https://www.google.com", True, False, "Not accessible"),
+        ("http:///a", False, True, "Not accessible"),
+        ("https://mate.academy", True, False, "Not accessible"),
+
+        ("https://www.youtube.com/watch?v=dQw4w9WgXcQ",
+         True, True, "Accessible"),
+
+        ("https://www.twitch.tv", True, True, "Accessible"),
+        ("https://gidhub.com", False, False, "Not accessible")
+    ],
```
YuriiKindrat, line 22: why do u need to use different urls here? 
could u use only one url and pass it directly to ur func ?

app/test_main.py
```diff
@@ -1 +1,46 @@
...
+@mock.patch("app.main.valid_google_url")
+@mock.patch("app.main.has_internet_connection")
+@pytest.mark.parametrize(
+    "url,is_valid,internet_connection,expected_status",
+    [
+        ("https://www.google.com", True, False, "Not accessible"),
+        ("http:///a", False, True, "Not accessible"),
+        ("https://mate.academy", True, False, "Not accessible"),
+
+        ("https://www.youtube.com/watch?v=dQw4w9WgXcQ",
```
YuriiKindrat, line 17: :)))

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #421, opened 2023-06-15
Solution

imelnyk007:


mariiahorbova (approved):

app/test_main.py
```diff
@@ -1 +1,33 @@
...
+def test_can_access_google_page(
+        connection_result: bool,
+        valid_result: bool,
+        expected_result: str
+) -> None:
+    with mock.patch("app.main.has_internet_connection") as mocked_connection:
+        with mock.patch("app.main.valid_google_url") as mocked_valid:
+            mocked_valid.return_value = valid_result
+            mocked_connection.return_value = connection_result
+            url = "https://www.google.com.ua"
```
YuriiKindrat, line 31: no need to create additional variable here

YuriiKindrat: u can pass url directly to the func

YuriiKindrat (changes requested):

app/test_main.py
```diff
@@ -1 +1,33 @@
...
+        "test_valid_and_connection_false"
+    ]
+)
+def test_can_access_google_page(
+        connection_result: bool,
+        valid_result: bool,
+        expected_result: str
+) -> None:
+    with mock.patch("app.main.has_internet_connection") as mocked_connection:
+        with mock.patch("app.main.valid_google_url") as mocked_valid:
```
YuriiKindrat, line 28: is it necessary to use `with` keyword twice here?

YuriiKindrat (changes requested):

Oleg-s-s (approved):
:+1:

PR #423, opened 2023-06-15
Tests "Can access google page"(First commit):

IhorVoskoboinikov:


Tarasidze (approved):
I'm going to give you a few suggestions.
If you use Parametrize and @mock.patch you will need only a single test function.
Just imagine that once somebody decides to change the URL or name of some callable functions in "main.py".  So you should refactor almost all lines in your code.

app/test_main.py
```diff
@@ -1 +1,39 @@
...
+from app.main import can_access_google_page
+
+
+@pytest.mark.parametrize(
+    "url, res_valid_google_url, res_internet_connection, expected",
+    [
+        ("https://www.google.com", True, True, "Accessible"),
+        ("https://www.google.com", True, False, "Not accessible"),
+        ("https://www.google.com", False, True, "Not accessible"),
+        ("https://www.google.com", False, False, "Not accessible"),
```
Oleg-s-s, line 14: Use a constant for url and then use it

Oleg-s-s (changes requested):

YuriiKindrat (approved):

PR #424, opened 2023-06-15
solution

mariiahorbova:


IhorVoskoboinikov (approved):
I also did something plausible, but received a response from colleagues that this is not what is needed))) Therefore, I have no additions to your code)))

app/test_main.py
```diff
@@ -1 +1,32 @@
-# write your code here
+from unittest.mock import patch
+from app.main import can_access_google_page
+
+
+def test_can_access_google_page() -> None:
+    with (patch("app.main.valid_google_url") as valid_url,
```
LiudmylaKulinchenko, line 6: Imagine having a fixture and not writing those managers again and again...

LiudmylaKulinchenko (changes requested):

app/test_main.py
```diff
@@ -1 +1,35 @@
...
+    action = can_access_google_page("https://www.google.com/")
+    assert action == "Not accessible"
+
+
+def test_no_access_if_only_url(manager_fixture: Type[fixture]) -> None:
+    valid_url, has_connection = manager_fixture
+    valid_url.return_value = True
+    has_connection.return_value = False
+    action = can_access_google_page("https://www.google.com/")
+    assert action == "Not accessible"
```
YuriiKindrat, line 35: could u use parametrize here to make ur code DRY ?
also what about False/False case ?

YuriiKindrat (changes requested):

app/test_main.py
```diff
@@ -1 +1,36 @@
...
+@pytest.mark.parametrize(
+    "url, connection, result",
+    [
+        (True, True, "Accessible"),
+        (False, True, "Not accessible"),
+        (True, False, "Not accessible"),
+        (False, False, "Not accessible")
+    ]
+)
+def test_can_access_google_page(
```
LiudmylaKulinchenko, line 26: You can just use mock.patch decorators to the tast instead of writing manager_fixture function with context manager
It will give the same affect and simplify the code

app/test_main.py
```diff
@@ -1 +1,36 @@
...
+        yield valid_url, has_connection
+
+
+@pytest.mark.parametrize(
+    "url, connection, result",
+    [
+        (True, True, "Accessible"),
+        (False, True, "Not accessible"),
+        (True, False, "Not accessible"),
+        (False, False, "Not accessible")
```
LiudmylaKulinchenko, line 23: Adding ids to your tests will allow for more descriptive test failure messages. Add them, please

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job! ☀🌺

PR #425, opened 2023-06-15
My tests

kovaliskoveronika:


app/test_main.py
```diff
@@ -1 +1,25 @@
...
+                         ids=[
+                             "url and connect True Accessible",
+                             "url False connect True Not accessible",
+                             "url True connect False Not accessible",
+                             "url and connect False Not accessible"
+                         ])
+def test_access_page(valid_url: bool,
+                     internet_connection:
+                     bool, answer: str) -> None:
+    with mock.patch("app.main.valid_google_url") as url, \
```
LiudmylaKulinchenko, line 21: Please, avoid using `\`, better use bracket

LiudmylaKulinchenko: Also you can use a decorator to make the code more clean

LiudmylaKulinchenko (changes requested):

YuriiKindrat (approved):

PR #426, opened 2023-06-15
Solution 1.01

Tarasidze:


app/test_main.py
```diff
@@ -1 +1,45 @@
...
+
+
+@fixture()
+def income_correct_url() -> str:
+    return "https://www.google.com/"
+
+
+@fixture()
+def income_incorrect_url() -> str:
+    return "https://www.chatgpt.com/"
```
Oleg-s-s, line 18: Fixtures are better to use if the URLs need to be fetched from external sources or some dynamically generated.
Here will be better to use some constant f.e.:
```
CORRECT_URL = "https://www.google.com/"
INCORRECT_URL = "https://www.chatgpt.com/"
```

Oleg-s-s (approved):

PR #428, opened 2023-06-18
Solution

volodymyr-komarnyckyi:


Mr-Furman (approved):
GJ

app/test_main.py
```diff
@@ -1 +1,42 @@
...
+            id="Connection denied because no internet connection"
+        ),
+        pytest.param(
+            False,
+            True,
+            "Not accessible",
+            id="Connection denied because url is not valid"
+        ),
+    ]
+)
```
Oleg-s-s, line 30: U should add a test for False, False

Oleg-s-s (changes requested):

PR #430, opened 2023-06-19
basic test

thorium-zero:


app/test_main.py
```diff
@@ -1 +1,23 @@
-# write your code here
+from unittest import mock
+import pytest
+from app.main import can_access_google_page
```
Kasatello, line 3: there should be an empty blank lane between built-in modules and local

Kasatello (approved):

app/test_main.py
```diff
@@ -1 +1,23 @@
...
+
+
+@pytest.mark.parametrize(
+    "url, has_internet, is_valid, expected",
+    [
+        ("https://www.google.com", True, True, "Accessible"),
+        ("https://www.google.com", False, True, "Not accessible"),
+        ("https://somewebsitename.com", True, False, "Not accessible"),
+        ("https://somewebsitename.com", False, False, "Not accessible")
```
YuriiKindrat, line 12: it seems to me there is no need to use different urls for this tests, so u can write an url only once directly in ur func

app/test_main.py
YuriiKindrat: agree, according to pep8 
https://peps.python.org/pep-0008/#module-level-dunder-names:~:text=Imports%20should%20be,group%20of%20imports.

app/test_main.py
```diff
@@ -1 +1,23 @@
...
+    ]
+)
+def test_can_access_google_page(
+    url: str, has_internet: bool, is_valid: bool, expected: str
+) -> None:
+    with mock.patch(
+            "app.main.has_internet_connection",
+            return_value=has_internet
+    ):
+        with mock.patch("app.main.valid_google_url", return_value=is_valid):
```
YuriiKindrat, line 22: no need to use `with` twice here, u can use it only once as for `open` function

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #433, opened 2023-06-19
Solution

Po-SvinSkipodla:


YuriiKindrat (changes requested):
could u use parametrize here? 
also what about False/False case ?))

YuriiKindrat (approved):

PR #437, opened 2023-06-20
Develop1

Kasatello:


app/test_main.py
```diff
@@ -1 +1,27 @@
...
+
+from app.main import can_access_google_page
+
+
+@pytest.mark.parametrize(
+    "url, expected_url, connection, expected_result",
+    [
+        pytest.param("https://www.google.com/", True, False, "Not accessible"),
+        pytest.param("https://www.google.com/", False, True, "Not accessible"),
+        pytest.param("https://www.google.com/", True, True, "Accessible")
```
YuriiKindrat, line 14: could u move ur urls directly to the func ? cause u are using the same url every time 
also what about False/False case ?

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #459, opened 2023-07-13
- solution

Oleksiy-Lyashenko:


app/test_main.py
```diff
@@ -1 +1,31 @@
...
+    can_access_google_page
+)
+
+
+@mock.patch("requests.get")
+def test_valid_google_url(mocked_request: any) -> None:
+    valid_google_url("")
+
+    mocked_request.assert_called_once()
+
```
pavlejviki, line 14: According to the task description you have to check only can_access_google_page functionality.

app/test_main.py
```diff
@@ -1 +1,31 @@
...
+
+@mock.patch("app.main.valid_google_url")
+def test_valid_google_url_works_in_can_access_google_page(
+        mocked_valid_url: any
+) -> None:
+    can_access_google_page("http://google.com")
+
+    mocked_valid_url.assert_called_once()
```
pavlejviki, line 22: You need to test the return value of the can_access_google_page function, mocking the values that the valid_google_url and has_internet_connection functions return.

Consider using @pytest.mark.parametrize.

pavlejviki (changes requested):

app/test_main.py
```diff
@@ -1 +1,99 @@
...
+
+@mock.patch("app.main.has_internet_connection")
+@pytest.mark.parametrize(
+    "initial_url,initial_hour,expected_result",
+    [
+        pytest.param(
+            "http://google.com",
+            True,
+            "Accessible",
+            id="check valid url and right hour"
```
pavlejviki, line 38: You need to mock the valid_google_url function in the same way as  the has_internet_connection and parametrize its return values to simulate different outcomes.

app/test_main.py
```diff
@@ -1 +1,99 @@
...
+    MissingSchema,
+    ConnectionError,
+    InvalidURL
+)
+
+from app.main import can_access_google_page
+
+
+@mock.patch("app.main.valid_google_url")
+def test_valid_google_url_works_in_can_access_google_page(
```
pavlejviki, line 13: The test naming  could be improved  test_has_internet_connection_is_called_in_can_access_google_page

app/test_main.py
```diff
@@ -1 +1,99 @@
...
+        mocked_valid_url: any
+) -> None:
+    can_access_google_page("http://google.com")
+
+    mocked_valid_url.assert_called_once()
+
+
+@mock.patch("app.main.has_internet_connection")
+def test_has_internet_connection_works_in_can_access_google_page(
```
pavlejviki, line 22: The test naming  could be improved  test_has_internet_connection_is_called_in_can_access_google_page

app/test_main.py
```diff
@@ -1 +1,99 @@
...
+        ),
+        pytest.param(
+            "",
+            MissingSchema,
+            id="empty sting is not a valid url"
+        ),
+
+    ]
+)
+def test_should_return_error_correctly(
```
pavlejviki, line 94: It would be more accurate to specify the name of the exact function being tested in the test naming.

pavlejviki (changes requested):

app/test_main.py
```diff
@@ -1 +1,99 @@
...
+            False,
+            "Not accessible"
+        )
+    ]
+)
+def test_can_access_google_page(
+        mocked_time: any,
+        initial_url: str,
+        initial_hour: int,
+        expected_result: bool
```
pavlejviki, line 57: Please correct the type annotations. mocked_time is a MagickMock object, initial_hour is a boolean value and the expected_result is a string

app/test_main.py
```diff
@@ -1 +1,99 @@
...
+            "",
+            MissingSchema,
+            id="empty sting is not a valid url"
+        ),
+
+    ]
+)
+def test_should_return_error_correctly(
+        initial_element: any,
+        expected_error: EOFError
```
pavlejviki, line 96: Fix type annotations here as well.

app/test_main.py
pavlejviki, line 94: This test seems invalid.

can_access_google_page only returns either "Accessible" or "Not accessible", but it doesn't raise any exceptions that you are trying to catch with pytest.raises(expected_error).

Additionally, you're using the raise keyword inappropriately in your test. You're trying to raise the result of can_access_google_page(initial_element), which will be a string, not an exception. The raise keyword in Python is used for raising exceptions, not for regular return values.

app/test_main.py
```diff
@@ -1 +1,98 @@
...
+        ),
+
+    ]
+)
+def test_should_return_error_when_ulr_invalid(
+        initial_element: any,
+        expected_error: any,
+) -> None:
+    with pytest.raises(expected_error):
+        raise valid_google_url(initial_element)
```
pavlejviki, line 98: Remove the raise keyword, just call the function.

pavlejviki (changes requested):

PR #465, opened 2023-07-23
Solution

tanya-burdenko:


app/test_main.py
```diff
@@ -1 +1,17 @@
...
+         mock.patch("app.main.has_internet_connection") as mocked_connection:
+
+        can_access_google_page(
+            "https://mate.academy/learn?module=python-core&course=python"
+        )
+
+        mocked_valid_url.assert_called_once_with(
+            "https://mate.academy/learn?module=python-core&course=python"
+        )
+        mocked_connection.assert_called_once()
```
ArsenPidhoretskyi, line 17: It is wrong realization. You must cover when some function return different data. True/False

ArsenPidhoretskyi (changes requested):

app/test_main.py
```diff
@@ -1 +1,29 @@
...
+        internet_connection: bool,
+        is_valid_url: bool,
+        outcome: str
+) -> None:
+    with mock.patch(
+            "app.main.has_internet_connection"
+    ) as mock_has_internet_connection,\
+        mock.patch(
+            "app.main.valid_google_url"
+    ) as mock_valid_url:
```
vsmutok, line 26: I think better to use decorator

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #469, opened 2023-07-25
solution

Davison322:


app/test_main.py
```diff
@@ -1 +1,10 @@
...
+from unittest import mock
+from app.main import can_access_google_page
+
+
+def test_functions_have_called() -> None:
+    with (mock.patch("app.main.valid_google_url") as mocked_valid,
+         mock.patch("app.main.has_internet_connection") as mocked_connect):
+        can_access_google_page("http://www.google.com")
+        mocked_valid.assert_called_once_with("http://www.google.com")
+        mocked_connect.assert_called_once()
```
ArsenPidhoretskyi, line 10: Add tests when functions return True/False

ArsenPidhoretskyi (changes requested):

PR #476, opened 2023-08-04
Solution

numerodiciannove:


app/test_main.py
```diff
@@ -1 +1,25 @@
...
+        expected: str,
+        monkeypatch: ModuleType
+) -> None:
+    monkeypatch.setattr(
+        "app.main.has_internet_connection",
+        lambda: internet_connection
+    )
+    monkeypatch.setattr(
+        "app.main.valid_google_url",
+        lambda url_: url_ == "https://www.google.com"
```
vasylhnatiuk, line 22: why did you use "https://www.google.com" here? you can mock only with True and False

app/test_main.py
```diff
@@ -1 +1,25 @@
-# write your code here
+from app.main import can_access_google_page
+import pytest
+from types import ModuleType
+
+
+@pytest.mark.parametrize("url, internet_connection, expected", [
+    ("https://www.google.com", True, "Accessible"),
```
vasylhnatiuk, line 7: write like 
    (True, True, "Accessible"),
    (False, False, "Not accessible"),
    (False, True, "Not accessible"),

app/test_main.py
```diff
@@ -1 +1,25 @@
...
+) -> None:
+    monkeypatch.setattr(
+        "app.main.has_internet_connection",
+        lambda: internet_connection
+    )
+    monkeypatch.setattr(
+        "app.main.valid_google_url",
+        lambda url_: url_ == "https://www.google.com"
+    )
+    result = can_access_google_page(url)
```
vasylhnatiuk, line 24: and here add link

vasylhnatiuk (changes requested):

vsmutok (approved):
Great 👍

PR #477, opened 2023-08-04
Develop

H1kki:


app/test_main.py
```diff
@@ -1 +1,19 @@
...
+    (True, True, "Accessible"),
+    (True, False, "Not accessible"),
+    (False, True, "Not accessible"),
+    (False, False, "Not accessible")
+])
+def test_can_access_google_page(internet_connection: bool,
+                                google_url: bool,
+                                expected_result: str) -> None:
+    with mock.patch("app.main.valid_google_url", return_value=google_url):
+        with mock.patch("app.main.has_internet_connection",
```
pavlejviki, line 16: you can use a single with statement to specify multiple context managers
```suggestion
with (
    mock.patch("app.main.valid_google_url", return_value=google_url),
    mock.patch("app.main.has_internet_connection", return_value=internet_connection)
):
```

app/test_main.py
```diff
@@ -1 +1,19 @@
...
+from unittest import mock
+from app.main import can_access_google_page
+
+
+@pytest.mark.parametrize("internet_connection, google_url, expected_result", [
+    (True, True, "Accessible"),
+    (True, False, "Not accessible"),
+    (False, True, "Not accessible"),
+    (False, False, "Not accessible")
+])
```
pavlejviki, line 11: Please use pytest.param and add identifiers to each test case.

pavlejviki (changes requested):

app/test_main.py
```diff
@@ -1 +1,33 @@
-# write your code here
+import pytest
+from unittest import mock
+from app.main import can_access_google_page
+
+
+@pytest.mark.parametrize("internet_connection, google_url, expected_result", [
+    pytest.param(True, True, "Accessible",
+                 id="should return Accessible"),
```
pavlejviki, line 8: When?

pavlejviki (approved):
Looks good, just fix the identifier for the first test case.

PR #479, opened 2023-08-05
Solution

nschcur:


app/test_main.py
```diff
@@ -1 +1,33 @@
...
+        monkeypatch: MonkeyPatch
+) -> None:
+    monkeypatch.setattr(main, "valid_google_url", lambda url: False)
+
+    url = "https://www.this-url-does-not-exist.com"
+    assert main.can_access_google_page(url) == "Not accessible"
+
+
+def test_can_access_google_page_with_unreachable_url() -> None:
+    url = "https://www.google.com/nonexistentpage"
```
vasylhnatiuk, line 23: use here also monkeypatch to mock valid_google_url and has_internet_connection

vasylhnatiuk (changes requested):

vsmutok (changes requested):
use @pytest.mark.parametrize

app/test_main.py
```diff
@@ -1 +1,22 @@
...
+def test_can_access_google_page(
+        url: str,
+        expected_result: str,
+        monkeypatch: MonkeyPatch
+) -> None:
+    monkeypatch.setattr(main, "has_internet_connection",
+                        lambda: False if "nonexistentpage" in url else True)
+
+    monkeypatch.setattr(main, "valid_google_url",
+                        lambda u: True if "google" in u else False)
```
vsmutok, line 20: I think better use  `mock.patch` then `MonkeyPatch`  in this case

vsmutok (changes requested):

app/test_main.py
```diff
@@ -1 +1,17 @@
...
+    ("https://www.google.com", "Accessible"),
+    ("https://www.this-url-does-not-exist.com", "Not accessible"),
+    ("https://www.google.com/nonexistentpage", "Not accessible")
+])
+def test_can_access_google_page(url: str, expected_result: str) -> None:
+    with mock.patch("app.main.has_internet_connection",
+                    return_value=True if "nonexistentpage"
+                                         not in url else False):
+        with mock.patch("app.main.valid_google_url",
+                        return_value=True if "google" in url else False):
```
ArsenPidhoretskyi, line 16: It is a bad realization. No need to pass url. Just pass as parametrize True/False values

ArsenPidhoretskyi (changes requested):

vsmutok (approved):
Good Job 👍

PR #484, opened 2023-08-14
Solution

Starchild666:


vasylhnatiuk (changes requested):
it would be better to use @pytest.mark.parametrize(

vsmutok (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #488, opened 2023-08-17
Solution

eduardhabryd:


Kroll410 (changes requested):
This is a good solution, but you can improve that with single test,but using pytest.mark.parameterize. The logic repeats in your 3 tests, so you can create parameter for different values and there will be 1 test with multiple runs :)

lgSilay (approved):
looks good!)

IvanRamyk (approved):

PR #490, opened 2023-08-19
Solution

OlPlaksa:


tests/test_tests.py
```diff
@@ -1,3 +1,5 @@
+from unittest import mock
+
```
vsmutok, line 2: why you need this import here?

vsmutok (changes requested):

vsmutok (approved):
Well Done

PR #491, opened 2023-08-19
resloved

VladyslavShepilov:


app/test_main.py
```diff
@@ -1 +1,30 @@
...
+        pytest.param(True, True,
+                     "Accessible",
+                     id="Has connection and valid url"),
+        pytest.param(False, True,
+                     "Not accessible",
+                     id="No connection but valid url"),
+        pytest.param(True, False,
+                     "Not accessible",
+                     id="Has connection invalid url")
+    )
```
vsmutok, line 20: what about False and False?

vsmutok (changes requested):
Goo

app/test_main.py
```diff
@@ -1 +1,33 @@
...
+        pytest.param(False, False,
+                     "Not accessible",
+                     id="No connection and not accessible")
+    )
+)
+def test_can_access_google_page(internet_connected: bool,
+                                valid_google_url: bool,
+                                can_access: str) -> None:
+    with mock.patch("app.main.has_internet_connection") as connection:
+        with mock.patch("app.main.valid_google_url") as url_validator:
```
ArsenPidhoretskyi, line 29: Use only one with context manager

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #494, opened 2023-08-23
Develop

nyanplague:


eduardhabryd (changes requested):
You have repetitions in your code. You can replace all those functions with one function along with pytest.mark.parametrize. Also, you can use monkeypatch to mock functions along with lambda functions and pass them directly into test function from pytest.mark.parametrize

nyanplague:
> You have repetitions in your code. You can replace all those functions with one function along with pytest.mark.parametrize. Also, you can use monkeypatch to mock functions along with lambda functions and pass them directly into test function from pytest.mark.parametrize


I have already removed "repetition" by using fixtures, so I believe it's just different way to do same thing.

eduardhabryd:
> > You have repetitions in your code. You can replace all those functions with one function along with pytest.mark.parametrize. Also, you can use monkeypatch to mock functions along with lambda functions and pass them directly into test function from pytest.mark.parametrize
> 
> I have already removed "repetition" by using fixtures, so I believe it's just different way to do same thing.

Not exactly, even if you want to use fixtures intead of monkeypatch you still have like 3 test functions that can be replaced with one + pytest.mark.parametrize

nyanplague:
> > > You have repetitions in your code. You can replace all those functions with one function along with pytest.mark.parametrize. Also, you can use monkeypatch to mock functions along with lambda functions and pass them directly into test function from pytest.mark.parametrize
> > 
> > 
> > I have already removed "repetition" by using fixtures, so I believe it's just different way to do same thing.
> 
> Not exactly, even if you want to use fixtures intead of monkeypatch you still have like 3 test functions that can be replaced with one + pytest.mark.parametrize

I tried your suggestion, and in my opinion it makes it much harder to read, so i decided to keep it and wait for mentor review. Thank you.

IvanRamyk (approved):

PR #499, opened 2023-08-26
Implemented tests

RPodolets:


OlegatorLE (changes requested):
I don't think you need to use a `class` here at all. You have never used `self `in your code.

OlegatorLE (approved):
it can be shorter, but ok )

IvanRamyk (changes requested):

IvanRamyk (approved):

PR #501, opened 2023-08-26
Solution

denys-source:


maxkatkalov (approved):
Logic of yr tests not so s1mple. We need mock and fixtures to make our life easier, but u chosen the hard way to use them. A more valuable approach here is to use parametrize and context manager for mocks + only one test. But yr tests work and it is good;)

PR #502, opened 2023-08-26
Implemented tests for can_access_google_page

KristianYe:


app/test_main.py
```diff
@@ -1 +1,39 @@
...
+            id="Access denied due to lack of internet",
+        ),
+        pytest.param(
+            False,
+            True,
+            "Not accessible",
+            id="Access denied due to unsatisfactory status code",
+        ),
+    ],
+)
```
innyshka, line 24: I think, one more test should be added when mock_url = False,  mock_internet = False

innyshka (changes requested):

app/test_main.py
KristianYe: I thought about it, but i'm not sure it's necessary, cause we already have two tests, which test both conditions separately, and one more test, which confirms that both conditions should be executed, but thank you for your advice)🙃

innyshka (approved):

app/test_main.py
IvanRamyk: Agree, it's better to add such test to be sure that the function works correctly

IvanRamyk (changes requested):

Kroll410 (approved):

PR #507, opened 2023-08-28
Emplemented tests of can_access_google_page function

ORENYT:
Used MonkeyPatch to mock functions

### DIFF
--- a/app/test_main.py
+++ b/app/test_main.py
@@ -1,1 +1,25 @@
-# write your code here
+import pytest
+from app.main import can_access_google_page
+
+
+@pytest.mark.parametrize(
+    "functions_result, response",
+    [
+        ([True, True], "Accessible"),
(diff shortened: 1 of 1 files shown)


app/test_main.py
```diff
@@ -1 +1,40 @@
...
+            "Accessible",
+            id="Should return 'Accessible' if both funcs return True",
+        ),
+        pytest.param(
+            [False, True],
+            "Not accessible",
+            id="Should return 'Not accessible' if invalid url",
+        ),
+        pytest.param(
+            [True, False],
```
IvanRamyk, line 19: I believe it's better to use two variables instead of a list to avoid "magic" indexes in the test body

IvanRamyk (changes requested):

Kroll410 (approved):

PR #517, opened 2023-08-28
google

VASO304:


app/test_main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+from unittest import mock
+import pytest
+from app.main import can_access_google_page
+
+
+@pytest.mark.parametrize("connection_status, valid_url, expected",
+                         [
+                             (True, True, "Accessible"),
```
katryana, line 8: Better to add  `pytest.param` > `id="..."` for better test recognition

katryana (changes requested):
Great Job!

app/test_main.py
VASO304: thanks for advise! did according to comment

Kroll410 (approved):

PR #528, opened 2023-09-03
Implemented access Google page

mtiunov:


app/test_main.py
```diff
@@ -1 +1,59 @@
-# write your code here
+import unittest
+from unittest.mock import patch
+from app.main import can_access_google_page
+
+
+class TestCanAccessGooglePage(unittest.TestCase):
```
EdAlekseiev, line 6: Please, rewrite your tests without using `unittest.TestCase` and with the help `pytest.mark.parametrize`

EdAlekseiev (changes requested):

Bohdan-Salamakha (approved):

PR #537, opened 2023-09-13
Solution

Artemh59:


app/test_main.py
```diff
@@ -1 +1,25 @@
...
+
+@mock.patch("app.main.valid_google_url")
+@mock.patch("app.main.has_internet_connection")
+def test_can_access_google_page(
+        mocked_has_internet_connection: bool,
+        mocked_valid_google_url: bool
+) -> None:
+
+    mocked_has_internet_connection.return_value = True
+    mocked_valid_google_url.return_value = True
```
pavlejviki, line 14: please use @pytest.mark.parametrize

pavlejviki (changes requested):

app/test_main.py
```diff
@@ -1 +1,43 @@
...
+            True,
+            False,
+            "Not accessible",
+            id="should return `Not accessible` when url is false"
+        )
+    ]
+)
+def test_can_access_google_page(
+        mocked_has_internet_connection: mock.MagicMock,
+        mocked_valid_google_url: mock.MagicMock,
```
pavlejviki, line 34: The order of mocks in your test function's parameters is incorrect relative to the order in which the @mock.patch decorators are applied.

pavlejviki (changes requested):

pavlejviki (approved):
Good job!

PR #538, opened 2023-09-14
can access google page

StepanShevchuk:


app/test_main.py
```diff
@@ -1 +1,33 @@
...
+    assert can_access_google_page("a") == "Not accessible"
+
+
+def test_cannot_access_if_only_valid_url(
+        mocked_valid_google_url: mock.MagicMock,
+        mocked_has_internet_connection: mock.MagicMock)\
+        -> AssertionError:
+    mocked_has_internet_connection.return_value = False
+    mocked_valid_google_url.return_value = True
+    assert can_access_google_page("a") == "Not accessible"
```
ArsenPidhoretskyi, line 33: Instead of duplicate code use pytest.parametize, also cover all possible options

ArsenPidhoretskyi (changes requested):

app/test_main.py
```diff
@@ -1 +1,33 @@
...
+def mocked_has_internet_connection() -> mock.MagicMock:
+    with mock.patch("app.main.has_internet_connection") as internet_connection:
+        yield internet_connection
+
+
+@pytest.mark.parametrize(
+    "valid_url,has_internet_connection,result",
+    [
+        (False, True, "Not accessible"),
+        (True, False, "Not accessible"),
```
ArsenPidhoretskyi, line 22: But what about True - True, False - False

ArsenPidhoretskyi (changes requested):

app/test_main.py
```diff
@@ -1 +1,35 @@
...
+)
+def test_cannot_access_if_only_one_true(
+        mocked_valid_google_url: mock.MagicMock,
+        mocked_has_internet_connection: mock.MagicMock,
+        valid_url: bool,
+        has_internet_connection: bool,
+        result: str) -> AssertionError:
+    mocked_has_internet_connection.return_value = has_internet_connection
+    mocked_valid_google_url.return_value = valid_url
+    assert can_access_google_page("a") == result
```
vasylhnatiuk, line 35:
```suggestion
    assert can_access_google_page("https//:test-link.com/") == result
```

vasylhnatiuk: it would be more readable

vasylhnatiuk (changes requested):

vsmutok (approved):
Well Done ✔️

PR #542, opened 2023-09-19
Solution

olgierrd:


app/test_main.py
```diff
@@ -1 +1,40 @@
...
+        has_internet_connection_mock: Callable
+) -> None:
+    valid_google_url_mock.return_value = True
+    has_internet_connection_mock.return_value = False
+    assert can_access_google_page("https://www.google.com") == "Not accessible"
+
+
+def test_not_accessible_invalid_url(valid_google_url_mock: Callable) -> None:
+    valid_google_url_mock.return_value = False
+    assert can_access_google_page("https://example.com") == "Not accessible"
```
Dimosphen1, line 40: There is a need to additionally test such cases as False - False, False - True. Besides that, you could use `parametrize` with test cases to avoid code duplication in your solution.

app/test_main.py
```diff
@@ -1 +1,40 @@
-# write your code here
+from typing import Callable
+
+import pytest
+from unittest.mock import patch
+from app.main import can_access_google_page
```
Dimosphen1, line 5: Order your imports in the correct way.

Dimosphen1 (changes requested):

app/test_main.py
```diff
@@ -1 +1,44 @@
-# write your code here
+import pytest
+from unittest.mock import patch
+from app.main import can_access_google_page
+from typing import Callable
```
viktoria-rybenchuk, line 4: fix order imports

viktoria-rybenchuk: Screenshot from 2023-09-19 20-47-04

viktoria-rybenchuk (changes requested):

app/test_main.py
```diff
@@ -1 +1,44 @@
...
+@pytest.fixture
+def valid_google_url_mock() -> None:
+    with patch("app.main.valid_google_url") as mock:
+        yield mock
+
+
+@pytest.fixture
+def has_internet_connection_mock() -> None:
+    with patch("app.main.has_internet_connection") as mock:
+        yield mock
```
pavlejviki, line 16: there is no need to create fixtures , you can simply use  the @patch decorator. It  is more straightforward and can be easier to read

pavlejviki (changes requested):

PR #554, opened 2023-10-05
'Solution'

tiron-vadym:


app/test_main.py
```diff
@@ -1 +1,67 @@
...
+        mock_valid_google_url: Mock,
+        mock_has_internet_connection: Mock
+) -> None:
+    mock_valid_google_url.return_value = False
+    mock_has_internet_connection.return_value = True
+
+    url = "https://www.invalidurl.com"
+    result = can_access_google_page(url)
+
+    assert result == "Not accessible"
```
Dimosphen1, line 54: There is no need to allocate a separate variable `result`, compare the result of the `can_access_google_page` call directly to a tested value.

app/test_main.py
```diff
@@ -1 +1,67 @@
...
+        mock_valid_google_url: Mock,
+        mock_has_internet_connection: Mock
+) -> None:
+    mock_valid_google_url.return_value = False
+    mock_has_internet_connection.return_value = False
+
+    url = "https://www.invalidurl.com"
+    result = can_access_google_page(url)
+
+    assert result == "Not accessible"
```
Dimosphen1, line 67: Use `pytest.mark.parametrize` in your solution to provide return values for `valid_google_url` and `has_internet_connection` and the expected result for `can_access_google_page` to avoid unnecessary code duplication.

Dimosphen1 (changes requested):

viktoria-rybenchuk (changes requested):
keep your test code clean, you can use a single decorator to patch multiple functions or methods `@mock.patch`

pavlejviki (changes requested):

app/test_main.py
```diff
@@ -1 +1,29 @@
...
+    "valid_url, has_internet, expected_result",
+    [
+        (True, True, "Accessible"),
+        (True, False, "Not accessible"),
+        (False, True, "Not accessible"),
+        (False, False, "Not accessible"),
+    ],
+)
+def test_can_access_google_page(
+        mock_has_internet_connection: Mock,
```
pavlejviki, line 20: fix the order of the parameters, should be the same as the decorators applied

pavlejviki (changes requested):

pavlejviki (approved):
Well done!

PR #558, opened 2023-10-07
'Solution'

nsdNite:


app/test_main.py
```diff
@@ -1 +1,29 @@
-# write your code here
+import pytest
+from app.main import can_access_google_page
+from unittest.mock import patch, MagicMock
```
viktoria-rybenchuk, line 3: fix imports ordering according to PEP8 (https://peps.python.org/pep-0008/#imports)

app/test_main.py
```diff
@@ -1 +1,29 @@
...
+
+
+@patch("app.main.valid_google_url")
+@patch("app.main.has_internet_connection")
+@pytest.mark.parametrize(
+    "connection_status, url_validation, expected_result",
+    [
+        (False, True, "Not accessible"),
+        (True, False, "Not accessible"),
+        (True, True, "Accessible")
```
viktoria-rybenchuk, line 13: What if we don't have internet connection and url is invalid?

viktoria-rybenchuk (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #561, opened 2023-10-11
solution

Depaton:


app/test_main.py
```diff
@@ -1 +1,28 @@
...
+    ("http://www.wed123.com/",
+     datetime.datetime(2023, 10, 10, 10, 0, 0),
+     "Not accessible"),
+    ("https://www.google.com/",
+     datetime.datetime(2023, 10, 10, 3, 0, 0),
+     "Not accessible"),
+    ("http://www.wed123.com/",
+     datetime.datetime(2023, 10, 10, 3, 0, 0),
+     "Not accessible")
+])
```
viktoria-rybenchuk, line 20: You can use  `True` and `False to represent the conditions, instead of hardcoding specific URLs and datetime.

viktoria-rybenchuk (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #564, opened 2023-10-13
'Solution'

lylakivan:


app/test_main.py
```diff
@@ -1 +1,29 @@
...
+    [
+        (False, False, "Not accessible"),
+        (False, True, "Not accessible"),
+        (True, False, "Not accessible"),
+        (True, True, "Accessible")
+    ]
+)
+def test_how_the_function_works(
+        mocked_connection: bool,
+        mocked_valid_url: bool,
```
pavlejviki, line 22: You should ensure that the parameter order in the test function definition matches the order that the decorators are applied

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #568, opened 2023-10-17
Solusion_2

Andru37:


app/test_main.py
```diff
@@ -1 +1,46 @@
...
+def test_can_access_google_page(
+        mocked_valid_google_url: mock,
+        mocked_has_internet_connection: bool,
+        validation_google_url: bool,
+        checking_has_internet_connection: bool,
+        return_can_access_page: str
+) -> None:
+    mocked_valid_google_url.return_value = validation_google_url
+    mocked_has_internet_connection.return_value =\
+        checking_has_internet_connection
```
EdAlekseiev, line 44: Use `()` instead of `\` for separating line

EdAlekseiev (changes requested):

app/test_main.py
Andru37: This is the Pycharm work.

EdAlekseiev (approved):
Good job

PR #570, opened 2023-10-19
Solution

mykolamateichuk:


Naz-iv (approved):
good job

app/test_main.py
```diff
@@ -1 +1,56 @@
-# write your code here
+from pytest import MonkeyPatch
+
+from app.main import can_access_google_page
+
+
+class TestCanAccessGooglePage:
```
fsocie7y, line 6: mb we can try to use parametrize or etc.?)
then code will be DRY imho

fsocie7y (changes requested):

app/test_main.py
mykolamateichuk: Yurii said that doing tests this way is better because you can run every test separately.

fsocie7y (approved):
ok)

PR #572, opened 2023-10-19
Solution

Naz-iv:


D1st3f (approved):
Could be written more succinctly, but in general everything is good

app/test_main.py
```diff
@@ -1 +1,49 @@
-# write your code here
+from unittest.mock import patch, MagicMock
+
+from app.main import can_access_google_page
+
+
```
fsocie7y, line 5: mb we can try to use parametrize or etc.?)
then code will be DRY imho

fsocie7y (changes requested):

app/test_main.py
Naz-iv: good suggestion, tried to rewrite it with parametrized test and mock functions with defined return value to make test easier to read

fsocie7y (approved):
u the best  🚀

PR #575, opened 2023-10-19
Solution

prochigor:


app/test_main.py
```diff
@@ -1 +1,52 @@
...
+        "should return Accessible when valid page and have connection",
+        "should return Not accessible when valid page, not have connection",
+        "should return Not accessible when not valid page, have connection",
+        "should return Not accessible when not valid page, not have connection"
+    ]
+)
+@mock.patch("app.main.valid_google_url")
+@mock.patch("app.main.has_internet_connection")
+def test_able_access_to_page_with(
+        mocked_valid_google_url: mock.MagicMock,
```
Familenko, line 43: First argument mocked_valid_google_url will take data from nearest decorator which is @mock.patch("app.main.has_internet_connection")

Familenko (changes requested):

app/test_main.py
prochigor: Thanks, I've figured it out in the next task)

fsocie7y (approved):
gj 🚀

PR #576, opened 2023-10-19
Develop

Familenko:


RVChornyy (commented):
I am not sure, but it seems, that you can enter allready valid booleans in @pytest.mark.parametrize. You are simulating valid_google_url and has_internet_connection functions, mocking them. Therefore mark parametrized data is considered to be a simulation of mocked function operation. Maybe I have mistaken, but you check your own inputs, testing yourself, when you use exceptions and time validation in test function.

Familenko:
> I am not sure, but it seems, that you can enter allready valid booleans in @pytest.mark.parametrize. You are simulating valid_google_url and has_internet_connection functions, mocking them. Therefore mark parametrized data is considered to be a simulation of mocked function operation. Maybe I have mistaken, but you check your own inputs, testing yourself, when you use exceptions and time validation in test function.

You are totally correct! I found just mocking is too simple and also it was requirement in description to install request, wich will be not in use in case of absolute mocking of the function. So I test all function. Most probably it will be rejected, because do not follow main task, but I try a lot of interesting things

katyakoziar (approved):

fsocie7y (approved):
well done ^_^

PR #591, opened 2023-10-21
Solution

skoval7:


fedorov-s-od (changes requested):
To make the code concise you should use @pytest.mark.parametrize

fsocie7y (approved):
GJ 🚀

PR #596, opened 2023-10-22
Solution Can access Google page

Oxbay:


Wonsky1 (approved):

app/test_main.py
```diff
@@ -1 +1,50 @@
-# write your code here
+from unittest import mock
+
+from app.main import can_access_google_page
+
+
+class TestCanAccessGooglePage:
+
```
fsocie7y, line 7: mb we can use parametrize?)

fsocie7y (changes requested):

app/test_main.py
Oxbay: No problemo)

app/test_main.py
```diff
@@ -1 +1,25 @@
...
+        (False, False, "Not accessible"),
+    ]
+)
+def test_valid_url_and_internet_connection(
+        valid_url: bool,
+        connection: bool,
+        get_access: str
+) -> None:
+    with mock.patch("app.main.valid_google_url") as mock_url, \
+            mock.patch("app.main.has_internet_connection") as mock_connection:
```
fsocie7y, line 22: u almost done)
pls dont use linebreakers `\`

fsocie7y: its bad practice

fsocie7y (changes requested):

app/test_main.py
Oxbay: ok

LiudmylaKulinchenko (approved):
Stunning! 💎

PR #597, opened 2023-10-22
'solution'

oleksandr-pupchenko:


MiskoRuslan (approved):
Nice, but you can try solve this tusk by using deco @pytest.mark.parametrize,to avoid code duplication and optimize its performance. 
~Then the process of "mocking" will be easier

app/test_main.py
```diff
@@ -1 +1,37 @@
-# write your code here
+from unittest.mock import patch
+
+from app.main import can_access_google_page
+
+
```
fsocie7y, line 5: please use parametrize

fsocie7y (changes requested):

tests/test_tests.py
```diff
@@ -1,54 +1,22 @@
+from unittest import mock
 import pytest
```
fsocie7y, line 2: u should do not change tests from mate) it was about ur own 😅

fsocie7y (changes requested):

tests/test_tests.py
oleksandr-pupchenko: oopsy :D

app/test_main.py
fsocie7y: pls do it in according to my recommedation))

fsocie7y (changes requested):

LiudmylaKulinchenko (approved):
Well done!

PR #600, opened 2023-10-23
Solution

Andrii-Silenko:


fedorov-s-od (approved):

app/test_main.py
```diff
@@ -1 +1,53 @@
...
+
+@pytest.fixture()
+def mocked_valid_google_url() -> Callable:
+    with mock.patch("app.main.valid_google_url") as mock_valid_google_url:
+        yield mock_valid_google_url
+
+
+@pytest.fixture()
+def mocked_has_internet_connection() -> Callable:
+    with mock.patch("app.main.has_internet_connection") as \
```
LiudmylaKulinchenko, line 15: please, avoid using \, better use brackets

app/test_main.py
```diff
@@ -1 +1,53 @@
-# write your code here
+from unittest import mock
+import pytest
+from app.main import can_access_google_page
+from typing import Callable
+
+
+@pytest.fixture()
+def mocked_valid_google_url() -> Callable:
+    with mock.patch("app.main.valid_google_url") as mock_valid_google_url:
```
LiudmylaKulinchenko, line 9: You can use decorators instead. And add a fixture to avoid writing them with every test

LiudmylaKulinchenko (changes requested):

PR #607, opened 2023-10-30
Solution

IvanMozhar:


app/test_main.py
```diff
@@ -1 +1,35 @@
...
+
+
+@mock.patch("app.main.valid_google_url", return_value=True)
+@mock.patch("app.main.has_internet_connection", return_value=False)
+def test_internet_connection_of_page(
+        valid_google_url: Callable,
+        has_internet_connection: Callable
+) -> None:
+    result = can_access_google_page("http://www.google.com")
+    assert result == "Not accessible"
```
Dimosphen1, line 35: Use `parametrize` from `pytest` to avoid code duplication and do not allocate the `result` variable, compare the function output directly in the assert statement. Besides that, add a test to check the output in case both `valid_google_url` and `has_internet_connection` return `False`.

app/test_main.py
```diff
@@ -1 +1,35 @@
-# write your code here
+from unittest import mock
+
+from app.main import can_access_google_page
+
+from typing import Callable
```
Dimosphen1, line 5: Order your imports in the correct way according to PEP-8.

Dimosphen1 (changes requested):

app/test_main.py
```diff
@@ -1 +1,55 @@
...
+    ]
+)
+def test_can_access_google_page(
+        url: str,
+        valid_google_url_value: bool,
+        has_internet_connection_value: bool,
+        expected: str
+) -> None:
+    with mock.patch("app.main.valid_google_url",
+                    return_value=valid_google_url_value), \
```
pavlejviki, line 52: use () instead of \

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #608, opened 2023-10-30
Done can access google page

OlenaLis123:


app/main.py
```diff
@@ -17,3 +17,5 @@ def can_access_google_page(url: str) -> str:
         return "Accessible"
     else:
         return "Not accessible"
+
+# print(can_access_google_page("http://seasonvar.ru/serial-11941-Tajny_Broukenvuda.html"))
```
EdAlekseiev, line 21: Delete this changes

app/test_main.py
```diff
@@ -1 +1,24 @@
-# write your code here
+from unittest import mock
+from app.main import can_access_google_page
+
+
+def test_called_function() -> None:
+    with (mock.patch("app.main.valid_google_url") as mock_valid,
+          mock.patch("app.main.has_internet_connection") as mock_internet):
```
EdAlekseiev, line 7: Use `mock.path` as decorator instead of context manager. It makes your code clearer.

app/test_main.py
```diff
@@ -1 +1,24 @@
...
+
+def test_accessible() -> None:
+    page = "http://seasonvar.ru/serial-11941-Tajny_Broukenvuda.html"
+    with (mock.patch("app.main.valid_google_url") as mock_valid,
+          mock.patch("app.main.has_internet_connection") as mock_internet):
+
+        mock_internet.return_value = True
+        mock_valid.return_value = True
+
+        assert can_access_google_page(page) == "Accessible"
```
EdAlekseiev, line 24: You also need to add additional test cases, when `mock_internet.return_value` returns `False` or  `mock_valid.return_value` returns `False`

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #615, opened 2023-11-06
solution_py-can-access-google-page_1

Vladkrit11:


app/test_main.py
```diff
@@ -1 +1,46 @@
...
+            False,
+            "Not accessible",
+            id="url isn't valid and there is no connection"
+        )
+    ],
+)
+def test_can_access_google_page(
+        valid_url: bool,
+        internet_connection: bool,
+        access: bool
```
Dimosphen1, line 40: `access` should not be a `bool`, it is a `string` from the incoming values. Change your type annotation to encounter that.

app/test_main.py
```diff
@@ -1 +1,46 @@
...
+) -> None:
+    with (mock.patch("app.main.valid_google_url") as mocked_url,
```
Dimosphen1, line 42: You could make two `mock.patch` decorators with functions and not create a context manager with each test, then you will only need to define the `return_value`.

Dimosphen1 (changes requested):

app/test_main.py
```diff
@@ -1 +1,46 @@
...
+        pytest.param(
+            False,
+            False,
+            "Not accessible",
+            id="url isn't valid and there is no connection"
+        )
+    ],
+)
+def test_can_access_google_page(
+        valid_url: str,
```
Dimosphen1, line 38: `valid_url` is a boolean value, change the type annotation for it to encounter the underlying type.

app/test_main.py
```diff
@@ -1 +1,46 @@
...
+        internet_connection: bool,
+        access: str
+) -> None:
+    with (mock.patch("app.main.valid_google_url",
+                     return_value=valid_url),
+          mock.patch("app.main.has_internet_connection",
+                     return_value=internet_connection)):
```
Dimosphen1, line 45: Create two `mock.patch` decorators with functions and not create a context manager with each test.

Dimosphen1 (changes requested):

app/test_main.py
```diff
@@ -1 +1,48 @@
 # write your code here
```
Dimosphen1, line 1: Do not forget to delete unnecessary comments in the future.

Dimosphen1 (approved):
LGTM, well done! Delete unnecessary comments, your code should contain only the required logic.

PR #620, opened 2023-11-11
Solution

viktor-codes:


app/test_main.py
```diff
@@ -1 +1,19 @@
...
+
+
+@mock.patch("app.main.has_internet_connection")
+@mock.patch("app.main.valid_google_url")
+def test_can_access_google_page(
+        mock_valid_google_url: mock.MagicMock,
+        mock_has_internet_connection: mock.MagicMock
+) -> None:
+    mock_has_internet_connection.return_value = True
+    mock_valid_google_url.return_value = True
```
ArsenPidhoretskyi, line 13: What about other cases, if this functions returns True/False False/True False/False

ArsenPidhoretskyi (changes requested):

rurakite:
I was thinking about that but was not sure if I could use pytest.mark.parametrize. Also, I was wondering about writing less code. Please check if it is OK now. And let me know is there any other way to do that better. Thank You

ArsenPidhoretskyi (approved):
Well done!

PR #622, opened 2023-11-13
'Solution'

Illya-Maznitskiy:


app/test_main.py
```diff
@@ -1 +1,31 @@
...
+
+@pytest.mark.parametrize(
+    "url, has_internet, is_valid, expected_result",
+    [
+        ("http://www.google.com", True, True, "Accessible"),
+        ("", False, True, "Not accessible"),
+        ("http://www.google.com", True, False, "Not accessible"),
+        ("", False, False, "Not accessible"),
+    ]
+)
```
viktoria-rybenchuk, line 15: You don't need to write these URLs here; you can just pass one URL in the can_access_google_page function. Since you already have mocked the value valid_url, it will work as expected.

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
Good job!

PR #630, opened 2023-11-24
Solution

CrazyHilly:


EdAlekseiev (approved):
Fix flake8

EdAlekseiev (changes requested):
Fix flake8, please

CrazyHilly:
Flake8 doesn't like single quotes where double quotes can be used. But those two lines that flake8 highlights are parts of previous sentences. Surely I should not use single quotes in one part of the sentence and double in the other, should I?
I replaced inner quotes with escape characters but, as always, flake8 doesn't like them either.
So I finally replaced the outer ones with triple quotes. 
Please let me know what is the best way, because I'm confused.

app/test_main.py
```diff
@@ -1 +1,39 @@
...
+    ("www.google.com", False, True, "Not accessible"),
+    ("www.google.com", False, False, "Not accessible"),
+]
+test_ids = [
+    """Should return "Accessible" if URL is valid and internet connection """
+    """is present""",
+    """Should return "Not accessible" if URL is invalid""",
+    """Should return "Not accessible" if internet connection is absent""",
+    """Should return "Not accessible" if URL is invalid and internet """
+    """connection is absent""",
```
EdAlekseiev, line 18:
```suggestion
    'Should return "Accessible" if URL is valid and internet connection '
    "is present",
    'Should return "Not accessible" if URL is invalid',
    'Should return "Not accessible" if internet connection is absent',
    'Should return "Not accessible" if URL is invalid and internet'
    "connection is absent",
```

EdAlekseiev (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #637, opened 2023-11-28
Solution

weuis:


app/test_main.py
```diff
@@ -1 +1,52 @@
...
+def test_can_access_google_page_not_accessible_no_internet_connection(
+        mock_valid_google_url: "patch",
+        mock_has_internet_connection: "patch"
+) -> None:
+    mock_valid_google_url.return_value = True
+    mock_has_internet_connection.return_value = False
+    url = "http://www.google.com"
+    result = can_access_google_page(url)
+
+    assert result == "Not accessible"
```
viktoria-rybenchuk, line 52: using  `@pytest.mark.parametrize` is generally considered more concise and readable.The use of `@mock.patch ` decorators helps isolate the function under test by mocking external dependencies.

viktoria-rybenchuk (changes requested):

app/test_main.py
```diff
@@ -1 +1,36 @@
...
+        valid_url: bool,
+        has_internet: bool,
+        expected_result: str
+) -> None:
+    mock_valid_google_url.return_value = valid_url
+    mock_has_internet_connection.return_value = has_internet
+
+    result = can_access_google_page(url)
+
+    assert result == expected_result
```
Dimosphen1, line 36: There is no need to allocate the variable `result`, you could evaluate the expression directly.

app/test_main.py
```diff
@@ -1 +1,36 @@
...
+@pytest.fixture
+def mock_valid_google_url() -> "patch":
+    with patch("app.main.valid_google_url") as mock:
+        yield mock
+
+
+@pytest.fixture
+def mock_has_internet_connection() -> "patch":
+    with patch("app.main.has_internet_connection") as mock:
+        yield mock
```
Dimosphen1, line 15: There is no need to create two separate functions, use  `@mock.patch` decorator and apply it to the `test_can_access_google_page` twice, e.g. `@mock.patch("app.main.valid_google_url")`. You could process it as is in the signature of the `test_can_access_google_page`.

app/test_main.py
```diff
@@ -1 +1,36 @@
...
+
+
+@pytest.mark.parametrize("url, valid_url, has_internet, expected_result", [
+    ("http://www.google.com", True, True, "Accessible"),
+    ("http://invalidurl", False, True, "Not accessible"),
+    ("http://www.google.com", True, False, "Not accessible"),
```
Dimosphen1, line 21: Add a test case when both `valid_url` and `has_internet` would be false.

app/test_main.py
```diff
@@ -1 +1,36 @@
...
+])
+def test_can_access_google_page(
+        mock_valid_google_url: "patch",
```
Dimosphen1, line 24: Use `mock.MagicMock` for the mock type annotation.

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #639, opened 2023-11-30
Created tests for can_access_google_page function.

AndriiHomeniuk1:


vasylhnatiuk (changes requested):
add test if all is False

PR #642, opened 2023-12-04
Develop

DenVia:


app/test_main.py
LiudmylaKulinchenko, line 1: Try to avoid code duplicating. You can use parametrize and not duplicate url in every test

LiudmylaKulinchenko (approved):
Great ☃

PR #643, opened 2023-12-04
Solution

lyaskov:


app/test_main.py
```diff
@@ -1 +1,30 @@
-# write your code here
+from unittest import mock
+
+from app.main import can_access_google_page
+
+
+def test_can_access_google_page_accessible() -> None:
+    with mock.patch("app.main.has_internet_connection") as h_i_c, \
+            mock.patch("app.main.valid_google_url") as v_g_u:
```
EdAlekseiev, line 8: Use `mock.patch` as decorator instead of context manager. Fix everywhere.

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #650, opened 2023-12-10
Solution

Liliyakhu:


app/test_main.py
```diff
@@ -1 +1,42 @@
...
+@mock.patch("app.main.has_internet_connection")
+def test_can_not_access_google_page_not_valid_site(
+        mocked_internet: callable,
+        mocked_validator: callable
+) -> None:
+    mocked_internet.return_value = True
+    mocked_validator.return_value = False
+    assert (
+        can_access_google_page("https://www.google.com/") == "Not accessible"
+    )
```
Dimosphen1, line 42: Do not duplicate your code, use `parametrize` from `pytest` with different combinations of `True` and `False` values for the mocked functions and the expected return text for the `can_access_google_page`. Besides that, add a test case when both `valid_google_url` and `has_internet_connection` return `False`.

Dimosphen1 (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #653, opened 2023-12-21
Solution

SuliganLP:


vasylhnatiuk (commented):
use @pytest.mark.parametrize to avoid duplicating

vasylhnatiuk (changes requested):
use @pytest.mark.parametrize to avoid duplicating

app/test_main.py
```diff
@@ -1 +1,44 @@
-# write your code here
+from unittest.mock import patch
+
+import pytest
+
+from app.main import can_access_google_page
+
+
+def test_valid_google_url() -> None:
```
vasylhnatiuk, line 8: why do you need it?

app/test_main.py
```diff
@@ -1 +1,44 @@
...
+    pass
+
+
+def test_has_internet_connection() -> None:
```
vasylhnatiuk, line 12: why do you need it?

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #655, opened 2023-12-25
Solution

rinotokun:


app/test_main.py
```diff
@@ -1 +1,24 @@
...
+    (False, True, "Not accessible"),
+    (False, False, "Not accessible"),
+])
+def test_can_access_google_page(
+        valid_url: bool,
+        internet_connection: bool,
+        expected_result: str
+) -> None:
+    with mock.patch("app.main.valid_google_url", return_value=valid_url):
+        with mock.patch(
```
pavlejviki, line 19: You can combine the two with statements into a single statement.

pavlejviki (changes requested):

app/test_main.py
```diff
@@ -1 +1,27 @@
...
+        internet_connection: bool,
+        expected_result: str
+) -> None:
+    with mock.patch.multiple(
+            "app.main",
+            valid_google_url=mock.Mock(return_value=valid_url),
+            has_internet_connection=mock.Mock(return_value=internet_connection)
+    ):
+        result = can_access_google_page("https://www.google.com")
+        assert result == expected_result
```
Dimosphen1, line 27: There is no need to create an additional variable, you could compare the result of the function directly to the expected result
```suggestion
        assert can_access_google_page("https://www.google.com") == expected_result
```

Do not create redundant variables in the future.

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration.

PR #659, opened 2024-01-02
Solution

gertsovenko:


app/test_main.py
```diff
@@ -1 +1,32 @@
...
+        expected_result: str
+) -> None:
+    with (
+        mock.patch("app.main.valid_google_url") as
+        mocked_valid_google_url
+    ):
+        with (
+            mock.patch("app.main.has_internet_connection") as
+            mocked_has_internet_connection
+        ):
```
Dimosphen1, line 27: Instead of creating two context managers one inside another, use a different approach with mocking functions with decorators using `mock.patch` and passing mock as a function argument, e.g. `@mock.patch("app.main.valid_google_url")`.

app/test_main.py
```diff
@@ -1 +1,32 @@
-# write your code here
+from unittest import mock
+import pytest
+from app.main import can_access_google_page
```
Dimosphen1, line 3: Add blank lines between different groups of imports according to PEP-8.

Dimosphen1 (changes requested):

app/test_main.py
```diff
@@ -1 +1,27 @@
-# write your code here
+from unittest import mock
+import pytest
```
Dimosphen1, line 2: `pytest` is a 3rd party test framework and `unittest` is a built-in python test framework, which is why you should separate them with one blank line as well. Take this into account in the future.

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration.

PR #661, opened 2024-01-05
Solution

Vityk-4W:


app/test_main.py
```diff
@@ -1 +1,44 @@
...
+import app.main
+
+
+def test_can_access_google_page_all_true() -> None:
+    with mock.patch(
+            "app.main.valid_google_url"
+    ) as mocked_valid_google_url:
+        with mock.patch(
+                "app.main.has_internet_connection"
+        ) as mocked_has_internet_connection:
```
EdAlekseiev, line 11: Use` mock.patch` as a decorator instead of context manager. Fix everywhere

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #672, opened 2024-01-15
'Solution' py-can-access-google-page

yura-push:


vasylhnatiuk (changes requested):
add test when accessible

app/test_main.py
```diff
@@ -1 +1,46 @@
...
+    assert can_access_google_page("https://mate.academy/") == "Not accessible"
+
+
+def test_can_access_google_page(
+        mock_valid_google_url: Mock,
+        mock_has_internet_connection: Mock
+) -> None:
+    mock_valid_google_url.return_value = True
+    mock_has_internet_connection.return_value = True
+    assert can_access_google_page("https://mate.academy/") == "Accessible"
```
pavlejviki, line 46: you can use pytest.mark.parametrize to parameterize your tests, which allows you to run the same test function with different sets of arguments.

pavlejviki (changes requested):

app/test_main.py
```diff
@@ -1 +1,45 @@
...
+    with patch("app.main.has_internet_connection") as mock_connection:
+        yield mock_connection
+
+
+@pytest.mark.parametrize(
+    "url, is_url_valid, has_connection, response",
+    [
+        ("https://mate.academy/", False, True, "Not accessible"),
+        ("https://mate.academy/", True, False, "Not accessible"),
+        ("https://mate.academy/", True, True, "Accessible")
```
pavlejviki, line 27: you should add a test case when is_url_valid, has_connection are both False

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #678, opened 2024-01-19
Solution

NataliiaKucher:


app/test_main.py
```diff
@@ -1 +1,15 @@
...
+
+
+@mock.patch("app.main.valid_google_url")
+@mock.patch("app.main.has_internet_connection")
+def test_can_access_google_page(mock_has_internet: None,
+                                mock_valid_google_url: None
+                                ) -> None:
+    mock_valid_google_url.return_value = True
+    mock_has_internet.return_value = True
+    assert can_access_google_page("http://example.com")
```
YukihiroSM, line 12: It's better to assert, that the function result is equal to something:
for example: 
```suggestion
    assert can_access_google_page("http://example.com") == "Success"
```

YukihiroSM (changes requested):
Can you please, add bit more test cases, like test, that one of parameters is False.

EdAlekseiev (approved):
Good job

PR #679, opened 2024-01-19
Solution

pnasonov:


vasylhnatiuk (changes requested):
add test when has_internet_connection and valid_google_url return False. and use @pytest.mark.parametrize(

vasylhnatiuk (approved):

PR #722, opened 2024-02-02
Solution

angelzagniy:


vasylhnatiuk (changes requested):
add test if mocked_valid_google_url and mocked_has_internet_connection are False, and aslo try to use @pytest.mark.parametrize

app/test_main.py
```diff
@@ -1 +1,54 @@
...
+def test_accessing_page_correctly(
+        url: str,
+        valid_google_url: bool,
+        internet_connection: bool,
+        expected_result: str
+) -> None:
+    with (mock.patch("app.main.valid_google_url")
+          as mocked_valid_google_url):
+        with (mock.patch("app.main.has_internet_connection")
+              as mocked_has_internet_connection):
```
pavlejviki, line 50: you can combine multiple context managers in a single with statement using commas to separate them.

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #744, opened 2024-02-15
solved task

Shantagust:


app/test_main.py
```diff
@@ -1 +1,37 @@
-# write your code here
+from unittest import mock
+from unittest import TestCase
```
Dimosphen1, line 2: You could import these two entities at once, no need to perform two imports

app/test_main.py
```diff
@@ -1 +1,37 @@
...
+from app.main import can_access_google_page
```
Dimosphen1, line 3: Add a blank line between different groups of imports according to PEP-8

app/test_main.py
```diff
@@ -1 +1,37 @@
...
+    @mock.patch("app.main.valid_google_url")
+    @mock.patch("app.main.has_internet_connection")
+    def test_return_result_when_not_has_internet_connection(
+            self,
+            mock_valid_google_url: mock,
+            mock_has_internet_connection: mock
+    ) -> None:
+        mock_valid_google_url.return_value = True
+        mock_has_internet_connection.return_value = False
+        self.assertEqual(can_access_google_page(""), "Not accessible")
```
Dimosphen1, line 37: Use a pytest parametrize decorator to define which arguments and what outcome you expect to obtain to not duplicate identical code three times. In this case, use of a class may be not needed. Besides that, add test case to test when both return values would be `False`.

Dimosphen1 (changes requested):

app/test_main.py
Shantagust: okay, thx for feedback

vsmutok (approved):
Good Job!

PR #745, opened 2024-02-16
Solution

Toltecos:


app/test_main.py
```diff
@@ -1 +1,33 @@
...
+    ]
+)
+def test_can_access_google_page(
+        valid_url: bool,
+        connection: bool,
+        access: str
+) -> None:
+    with (
+        mock.patch("app.main.valid_google_url") as mock_valid_google_url,
+        mock.patch("app.main.has_internet_connection") as mock_has_connection
```
Dimosphen1, line 29: You could use `mock.patch` as decorators as well and pass them to the function as arguments.

app/test_main.py
```diff
@@ -1 +1,33 @@
-# write your code here
+import pytest
+from unittest import mock
```
Dimosphen1, line 2: `unittest` is a built-in module, while `pytest` is not. Pay closer attention to the imports next time.

Dimosphen1 (approved):
LGTM, well done! Several changes were left for consideration.

PR #747, opened 2024-02-16
added solution

MAFIOZNIK1999:


app/test_main.py
```diff
@@ -1 +1,55 @@
...
+
+
+@mock.patch("app.main.valid_google_url")
+@mock.patch("app.main.has_internet_connection")
+def test_when_url_is_correct_and_has_connection(
+        mocked_valid_google_url: bool,
+        mocked_internet_connection: bool
+) -> None:
+    mocked_internet_connection.return_value = True
+    mocked_valid_google_url.return_value = True
```
LiudmylaKulinchenko, line 13: You can also use parametrize here, since you have similar tests logic

app/test_main.py
```diff
@@ -1 +1,55 @@
...
+
+    assert can_access_google_page("https://google.com") == "Accessible"
+
+
+@mock.patch("app.main.valid_google_url")
+@mock.patch("app.main.has_internet_connection")
+def test_when_url_is_incorrect(
```
LiudmylaKulinchenko, line 20: Imagine having a fixture and not writing those decorators again and again..

LiudmylaKulinchenko (changes requested):

PR #750, opened 2024-02-18
py-can-access-google-page

darkeliss:


vasylhnatiuk (changes requested):
add test if, and avoid duplicaing of code 
```
mocked_internet.return_value = False
mocked_validator.return_value = False
```

EdAlekseiev (approved):
Good job

PR #751, opened 2024-02-18
implemented tests for func can_access_google_page w/ mock

RomanHlodann:


app/test_main.py
```diff
@@ -1 +1,47 @@
...
+
+import pytest
+
+from app.main import can_access_google_page
+
+
+@pytest.fixture
+def mock_valid_google_url() -> mock.MagicMock:
+    with mock.patch("app.main.valid_google_url") as mocked_url:
+        mocked_url.return_value = True
```
Dimosphen1, line 11: Do not define a default return value in mock, you should define it in a test itself.

app/test_main.py
```diff
@@ -1 +1,47 @@
...
+def test_return_not_accessible(
+        mock_valid_google_url: mock.MagicMock,
+        mock_has_internet_connection: mock.MagicMock
+) -> None:
+    mock_valid_google_url.return_value = False
+    assert can_access_google_page("/") == "Not accessible"
+
+    mock_valid_google_url.return_value = True
+    mock_has_internet_connection.return_value = False
+    assert can_access_google_page("/") == "Not accessible"
```
Dimosphen1, line 47: Reformat your code in a way that would not use code duplication to run similar checks. You could use `parameterize` from `pytest` to test different combinations for `mock_valid_google_url` and `mock_has_internet_connection` and the `can_access_google_page` output such as (True, True, "Accessible").

Dimosphen1 (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #757, opened 2024-02-25
Solution

MykhailoIvchenko:


app/test_main.py
```diff
@@ -1 +1,67 @@
...
+
+
+def test_has_internet_connection_has_been_called(
+        mocked_has_internet_connection: Any) -> Any:
+    app.main.can_access_google_page(url)
+    mocked_has_internet_connection.assert_called_once()
+
+
+def test_valid_google_url_has_been_called_with_url(
+        mocked_valid_google_url: Any) -> Any:
```
vasylhnatiuk, line 32:
```suggestion
        mocked_valid_google_url: Callable) -> None:
```

vasylhnatiuk: fix annotation everywhere

vasylhnatiuk (changes requested):

app/test_main.py
MykhailoIvchenko: Done

EdAlekseiev (approved):
Good job

PR #765, opened 2024-03-03
Solution

kolya2731:


app/test_main.py
```diff
@@ -1 +1,28 @@
-# write your code here
+from app.main import can_access_google_page
+
+from unittest import mock
+
+import pytest
```
Dimosphen1, line 5: Order your imports according to PEP-8

Dimosphen1 (changes requested):

app/test_main.py
```diff
@@ -1 +1,28 @@
-# write your code here
+import pytest
+
+from unittest.mock import MagicMock, patch
+
+from app.main import can_access_google_page
```
Dimosphen1, line 5: Your imports are not ordered correctly. As mentioned in the [PEP-8](https://peps.python.org/pep-0008/#imports), imports should have the following ordering:

1. Standard library imports.
2. Related third party imports.
3. Local application/library specific imports.
```suggestion
from unittest.mock import MagicMock, patch

import pytest

from app.main import can_access_google_page

```

Dimosphen1 (changes requested):

PR #781, opened 2024-03-16
Solution

ILyakhova:


NazarSheremet (changes requested):
In this task better to use `@pytest.mark.parametrize`. This will reduce code duplication.

PR #784, opened 2024-03-18
'Solution'

ArtomUmantsev:


app/test_main.py
```diff
@@ -1 +1,22 @@
...
+
+    mock_valid_google_url.return_value = False
+    assert can_access_google_page("/www.invalidurl.com") == "Not accessible"
+
+    mock_has_internet_connection.return_value = False
+    mock_valid_google_url.return_value = True
+    assert can_access_google_page("www.google.com") == "Not accessible"
+
+    mock_valid_google_url.return_value = False
+    assert can_access_google_page("www.invalidurl.com") == "Not accessible"
```
EdAlekseiev, line 22: Rewrite your test case with using `parametrize`
https://docs.pytest.org/en/7.1.x/example/parametrize.html

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #785, opened 2024-03-18
Solution

Viesich:


app/test_main.py
```diff
@@ -1 +1,47 @@
-# write your code here
+from unittest import mock
```
Dimosphen1, line 1: Add a blank line between different groups of imports according to PEP-8

app/test_main.py
```diff
@@ -1 +1,47 @@
...
+
+def test_can_access_google_page_no_internet_connection() -> None:
+    with (
+        mock.patch("app.main.has_internet_connection")
+        as mock_has_internet_connection
+    ):
+        mock_has_internet_connection.return_value = False
+        assert can_access_google_page(
+            "https://www.google.com"
+        ) == "Not accessible"
```
Dimosphen1, line 47: Create one function and use two boolean values for mocks and an expected result as a parameter that is passed. Your first parameter to the `@pytest.mark.parametrize` should be like `"valid_url_value, connection_value, expected_value"`

Dimosphen1 (changes requested):

PR #786, opened 2024-03-19
Solution

nicksetrakov:


app/test_main.py
```diff
@@ -1 +1,39 @@
...
+
+@pytest.mark.parametrize(
+    "valid_url,internet_connection,expected_result",
+    [
+        (True, True, "Accessible"),
+        (True, False, "Not accessible"),
+        (False, True, "Not accessible"),
+        (False, False, "Not accessible"),
+    ]
+)
```
YuriiKindrat, line 28: it is better to add ids here

YuriiKindrat (changes requested):
LGTM

app/test_main.py
```diff
@@ -1 +1,45 @@
...
+@patch("app.main.valid_google_url")
+@patch("app.main.has_internet_connection")
+def test_should_return_expected_result(
+        mock_has_internet_connection: MagicMock,
+        mock_valid_google_url: MagicMock,
+        valid_url: bool,
+        internet_connection: bool,
+        expected_result: str) -> None:
```
YuriiKindrat, line 42: also it is bad formatting here, didn't notice this before (

YuriiKindrat (changes requested):

app/test_main.py
nicksetrakov: I'm too)

Oleksl888 (approved):
good job!

PR #792, opened 2024-03-21
Solution

Bujhm99:


app/test_main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+from typing import Callable
+from unittest import mock
+
+from app.main import can_access_google_page
+
+
+@mock.patch("app.main.has_internet_connection")
+def test_has_internet_connection_call(mocked: Callable) -> None:
```
nicksetrakov, line 8: better use for mock annotation - MagicMock

app/test_main.py
```diff
@@ -1 +1,16 @@
...
+    can_access_google_page("https://www.google.com/webhp?client=firefox-b-d")
+    mocked.assert_called_once()
+
+
+@mock.patch("app.main.valid_google_url")
+def test_have_valid_google_url(mocked: Callable) -> None:
+    can_access_google_page("https://www.google.com/webhp?client=firefox-b-d")
+    mocked.assert_called_once()
```
nicksetrakov, line 16: And here valid_google_url accepts argument so better called assert_called_once_with

nicksetrakov (changes requested):

nicksetrakov (approved):

app/test_main.py
```diff
@@ -1 +1,17 @@
...
+
+
+@mock.patch("app.main.valid_google_url")
+def test_have_valid_google_url(mocked: mock.MagicMock) -> None:
+    can_access_google_page("https://www.google.com/webhp?client=firefox-b-d")
+    mocked.assert_called_once_with(
+        "https://www.google.com/webhp?client=firefox-b-d"
+    )
```
YuriiKindrat, line 17: what about testing combination of returns for mocked func ?)

YuriiKindrat (changes requested):

app/test_main.py
```diff
@@ -1 +1,43 @@
...
+    ]
+)
+def test_combination_of_returns(has_connect: bool,
+                                valid_page: bool,
+                                result: str) -> None:
+    with (mock.patch("app.main.has_internet_connection") as mock_connect,
+          mock.patch("app.main.valid_google_url") as mock_valid):
+        mock_connect.return_value = has_connect
+        mock_valid.return_value = valid_page
+        link = "https://www.google.com/webhp?client=firefox-b-d"
```
Oleksl888, line 42: link must be  placed in a fixture since you reuse it in other tests

app/test_main.py
```diff
@@ -1 +1,43 @@
...
+        "All is fine",
+        "Wrong page",
+        "No connection",
+        "No connection and Wrong page"
+    ]
+)
+def test_combination_of_returns(has_connect: bool,
+                                valid_page: bool,
+                                result: str) -> None:
+    with (mock.patch("app.main.has_internet_connection") as mock_connect,
```
Oleksl888, line 38: you can use decorators instead to make the code appear cleaner

app/test_main.py
```diff
@@ -1 +1,43 @@
...
+import pytest
+
+
+from app.main import can_access_google_page
+
+
+@mock.patch("app.main.has_internet_connection")
+def test_has_internet_connection_call(mocked: mock.MagicMock) -> None:
+    can_access_google_page("https://www.google.com/webhp?client=firefox-b-d")
+    mocked.assert_called_once()
```
Oleksl888, line 11: you should add a descriptive message here in case if test fails - same in function below

app/test_main.py
```diff
@@ -1 +1,43 @@
...
+
+@mock.patch("app.main.valid_google_url")
+def test_have_valid_google_url(mocked: mock.MagicMock) -> None:
+    can_access_google_page("https://www.google.com/webhp?client=firefox-b-d")
+    mocked.assert_called_once_with(
+        "https://www.google.com/webhp?client=firefox-b-d"
+    )
+
+
+@pytest.mark.parametrize(
```
Oleksl888, line 22: What about test for rasing excpetions when wrong data is passed

Oleksl888 (changes requested):

app/test_main.py
Bujhm99: We can do mock for datetime.datetime.now() and pass wrong format of date and get AttributeError, but our main function was created without taking into account of this exeption.

YuriiKindrat (approved):
GJ

PR #796, opened 2024-03-23
Solution

AndriiDraga:


app/test_main.py
```diff
@@ -1 +1,63 @@
...
+)
+def test_cannot_access_if_only_valid_url(
+        monkeypatch: MonkeyPatch,
+        url: str,
+        expected_status: str
+) -> None:
+    monkeypatch.setattr(
+        "app.main.valid_google_url", lambda url: False
+    )
+    assert can_access_google_page(url) == expected_status
```
Dimosphen1, line 63: Pass mocks as decorators to a function using `@mock.patch`. Create one function and use two boolean values for mocks and an expected result as a parameter that is passed. Your first parameter to the ` @pytest.mark.parametrize` should be like `"valid_url_value, connection_value, expected_value"`.

app/test_main.py
```diff
@@ -1 +1,63 @@
-# write your code here
+import pytest
+
```
Dimosphen1, line 2: There is no need to add blank line here

Dimosphen1 (changes requested):

app/test_main.py
```diff
@@ -1 +1,30 @@
...
+        expected_value: str
+) -> None:
+    with (
+        mock.patch(
+            "app.main.valid_google_url",
+            return_value=valid_url_value
+        ),
+        mock.patch(
+            "app.main.has_internet_connection",
+            return_value=connection_value)):
```
Dimosphen1, line 28: Apply mocks as decorators and pass them to the function instead to avoid creating a context manager got the whole body of the function

app/test_main.py
```diff
@@ -1 +1,30 @@
-# write your code here
+import pytest
+from unittest import mock
+
+from app.main import can_access_google_page
```
Dimosphen1, line 4: Add a blank line between different import groups and order them correctly to follow PEP-8

Dimosphen1 (changes requested):

app/test_main.py
Dimosphen1: This comment is not implemented in two change requests in line. In case you face errors, please visit Q&A to discuss it with mentor. In other case, please apply two mocks as two function decorators to the `test_access_google_page`.

Dimosphen1 (changes requested):

pavlopro (approved):
Good Job!

PR #800, opened 2024-03-26
Pochtalon

pochtalon:


app/test_main.py
```diff
@@ -1 +1,31 @@
...
+from app.main import can_access_google_page
+
+
+url = "https://google.com"
+
+
+@pytest.mark.parametrize(
+    "valid_url, has_internet, can_access",
+    [
+        (True, 7, "Accessible"),
```
NazarSheremet, line 15: I can't understand the logic of this test case. Why the value of has_internet is int?

NazarSheremet (changes requested):
Need changes or explanation

app/test_main.py
pochtalon: In the very begining I made mistake and put to the parameters valid/invalid time. Then fixed it, but, I don't know why, only for 3 last cases)

YuriiSmolii (approved):
Well done.

PR #807, opened 2024-03-28
'Solution'

o-zinevych:


app/test_main.py
```diff
@@ -1 +1,15 @@
...
+    with (
+        mock.patch("app.main.has_internet_connection") as mock_connection,
+        mock.patch("app.main.valid_google_url") as mock_valid_url
+    ):
+        url = ("https://www.google.com/webhp?hl=en&sa=X&ved="
+               "0ahUKEwjlr5aw25eFAxVFFBAIHbjlAdEQPAgJ")
+        can_access_google_page(url)
+
+        mock_connection.assert_called_once()
+        mock_valid_url.assert_called_once_with(url)
```
EdAlekseiev, line 15: You should check whether `can_access_google_page` returns the correct value in different cases

EdAlekseiev (changes requested):

vsmutok (changes requested):
DRY 

so many code duplications
use `pytest.mark.parametrize`

PR #810, opened 2024-04-01
'Solution'

PozniakVA:


app/test_main.py
```diff
@@ -1 +1,32 @@
-# write your code here
+import pytest
+
+from unittest import mock
+
+from unittest.mock import MagicMock
+
+from app.main import can_access_google_page
```
Dimosphen1, line 7: Add one blank line between different import groups and order them correctly to follow PEP-8

app/test_main.py
```diff
@@ -1 +1,32 @@
...
+
+
+@pytest.mark.parametrize(
+    "valid_google_url,has_internet_connection, expected",
+    [
+        (True, True, "Accessible"),
+        (False, False, "Not accessible"),
+        (True, False, "Not accessible"),
+        (False, True, "Not accessible"),
+        (False, False, "Not accessible")
```
Dimosphen1, line 17: There is no need to duplicate test cases, they should be unique

app/test_main.py
```diff
@@ -1 +1,32 @@
...
+        mocked_url: MagicMock,
+        mocked_internet_connection: MagicMock,
+        valid_google_url: bool,
+        has_internet_connection: bool,
+        expected: str
+) -> None:
+    mocked_url.return_value = valid_google_url
+    mocked_internet_connection.return_value = has_internet_connection
+    result = can_access_google_page("url")
+    assert result == expected
```
Dimosphen1, line 32: There is no need for the `result` variable, you could evaluate `can_access_google_page("url")` directly

Dimosphen1 (changes requested):

vasylhnatiuk (approved):

PR #811, opened 2024-04-02
Develop

March1205:


vasylhnatiuk (changes requested):
use @pytest.mark.parametrize to avoid duplicating

vasylhnatiuk (changes requested):
also add test for ("https://www.google.com", False, False, "Not accessible")

vasylhnatiuk (approved):

PR #812, opened 2024-04-02
Solution

vitalinamal:


app/test_main.py
```diff
@@ -1 +1,28 @@
...
+
+
+@pytest.mark.parametrize(
+    "valid_url, has_connection, expected_output",
+    [
+        (True, True, "Accessible"),
+        (True, False, "Not accessible"),
+        (False, True, "Not accessible"),
+        (False, False, "Not accessible"),
+    ]
```
Oleksl888, line 14: what about id/ids here?

app/test_main.py
```diff
@@ -1 +1,28 @@
-# write your code here
+import pytest
+
+from unittest.mock import MagicMock, patch
```
Oleksl888, line 3: check import orders here

app/test_main.py
```diff
@@ -1 +1,28 @@
...
+        mock_valid_google_url: MagicMock,
+        mock_has_internet_connection: MagicMock,
+        valid_url: bool,
+        has_connection: bool,
+        expected_output: str
+) -> None:
+    mock_valid_google_url.return_value = valid_url
+    mock_has_internet_connection.return_value = has_connection
+
+    assert can_access_google_page("https://www.google.com/") == expected_output
```
Oleksl888, line 28: don't you need to check that your mocked functions has been called as well?

Oleksl888 (changes requested):

YuriiKindrat (approved):
GJ

PR #814, opened 2024-04-02
Solution

panicua:
I got to admit that it was one of the hardest tasks for the past 2 weeks for me, I wouldn't probably make it using only platform material, had to use some external sources, one of them were the video on youtube where the guy were doing some similar stuff for one of the functions (url response mocking) ```youtube.com/watch?v=-F6wVOlsEAM&ab_channel=NeuralNine```, the other one is our well known OpenAI.

I decided to leave 2 comments for you to easier to understand my code, eventhough it's said to clean everything before commit in ```checklist.md```

P.S. I'll never disown people who write tests on regular basics ;D

### DIFF
--- a/app/test_main.py
+++ b/app/test_main.py
@@ -28,14 +28,10 @@ def test_can_access_google_page(mock_valid_google_url,
     mock_valid_google_url.return_value = is_valid_url
     mock_has_internet_connection.return_value = has_connection
 
-    can_access_google_page("https://somefakegoogle.net/")
+    can_access_google_page("https://smfakeggl.net/")
 
     mock_has_internet_connection.assert_called_once()
-    if mock_has_internet_connection.return_value:
-        mock_valid_google_url.assert_called_once_with(
-            "https://somefakegoogle.net/"
-        )
+    if mock_has_internet_connection.assert_called_once():
+        mock_valid_google_url.assert_called_once_with("https://smfakeggl.net/")
 
-    assert can_access_google_page(
-        "https://somefakegoogle.net/"
-    ) == expected_result
+    assert can_access_google_page("https://smfakeggl.net/") == expected_result


app/test_main.py
```diff
@@ -1 +1,28 @@
-# write your code here
+import unittest
+from unittest import mock
+
+from app.main import can_access_google_page
+
+
+class TestCanAccessGooglePage(unittest.TestCase):
+    @mock.patch("requests.get")
+    @mock.patch("datetime.datetime")
```
YuriiKindrat, line 9: i know it is quite hard to mock datetime function, but the task was to mock other funcs
image
it seems to me, that task would be much easier for u, if u choose this funcs to mock)
after mocking this funcs, u could test the work of function changing their return values (u could use parametrize for this)

YuriiKindrat (changes requested):

app/test_main.py
panicua: Well, in this case if I understood you right it's much easier to do than what I was trying to do :)
I'll try to fix it after the morning lesson

panicua:
removed mocks of requests, datetime and their methods

panicua:
I added to flake8 to ignore ```ANN001``` for this task cause I don't think annotation is even needed for tests, like ```mock_connection: mock.MagicMock``` is pretty useless, isn't it? and lecturer Yurii said that it's fine during the lesson.

I had to remove class at all ```class TestCanAccessGooglePage(unittest.TestCase):``` cause it didn't work well for me when I tried to use decorators of pytest/unittest under the unittest class maybe some conflicts with those modules, it was giving me errors with positional attributes but even chat GPT didn't understand why :D so I decided to make it only function with decorators.

Overall, this task is hell

YuriiKindrat (approved):
GJ

PR #817, opened 2024-04-03
Solution

Gilgumish:


app/test_main.py
```diff
@@ -1 +1,29 @@
...
+
+@pytest.mark.parametrize(
+    "internet_connection, url_valid, expected_result",
+    [
+        (True, True, "Accessible"),
+        (False, True, "Not accessible"),
+        (True, False, "Not accessible"),
+        (False, False, "Not accessible"),
+    ]
+)
```
YuriiKindrat, line 16: it is common practice to ad ids to parametrize

YuriiKindrat (changes requested):

YuriiKindrat (approved):
GJ

PR #819, opened 2024-04-03
Solution

DmytroHlazyrin:


app/test_main.py
```diff
@@ -1 +1,35 @@
...
+
+@pytest.mark.parametrize(
+    "valid_google_url_return, has_internet_connection_return, expected_result",
+    [
+        (True, True, "Accessible"),
+        (True, False, "Not accessible"),
+        (False, True, "Not accessible"),
+        (False, False, "Not accessible")
+    ]
+)
```
YuriiKindrat, line 16: it is common practice to add ids to parametrize

YuriiKindrat (changes requested):

app/test_main.py
```diff
@@ -1 +1,41 @@
-# write your code here
+import pytest
+from unittest import mock
```
Oleksl888, line 2: be consistent with order of imports

Oleksl888 (approved):

PR #820, opened 2024-04-03
Solution

y-kondrashova:


app/test_main.py
```diff
@@ -1 +1,44 @@
...
+
+    @patch("app.main.valid_google_url", return_value=False)
+    @patch("app.main.has_internet_connection", return_value=False)
+    def test_invalid_url_and_no_internet_connection(
+            self,
+            mock_valid_google_url: bool,
+            mock_has_internet_connection: bool
+    ) -> None:
+        result = can_access_google_page("htt//ww.someting_wrong.com")
+        self.assertEqual(result, "Not accessible")
```
YuriiKindrat, line 44: could u use parametrize here to optimize ur code ?

YuriiKindrat (changes requested):

app/test_main.py
```diff
@@ -1 +1,32 @@
...
+        pytest.param(False, True, "Not accessible", id="You have wrong url"),
+        pytest.param(False,
+                     False,
+                     "Not accessible",
+                     id="There is no internet connection and wrong url")
+    ]
+)
+@patch("app.main.valid_google_url")
+@patch("app.main.has_internet_connection")
+def test_can_access_google_page(mocked_valid_google_url: bool,
```
Oleksl888, line 24: wrong type on mocked functions

app/test_main.py
```diff
@@ -1 +1,32 @@
-# write your code here
+import pytest
+from unittest.mock import patch
```
Oleksl888, line 2: check order of imports here, built in libraries come first

app/test_main.py
```diff
@@ -1 +1,32 @@
...
+@patch("app.main.has_internet_connection")
+def test_can_access_google_page(mocked_valid_google_url: bool,
+                                mocked_has_internet_connection: bool,
+                                is_valid_url: bool,
+                                is_internet_connection: bool,
+                                expected_result: str) -> None:
+    mocked_valid_google_url.return_value = is_valid_url
+    mocked_has_internet_connection.return_value = is_internet_connection
+
+    assert can_access_google_page("https://www.google.com") == expected_result
```
Oleksl888, line 32: you should also check that your mocked functions actually worked

Oleksl888 (changes requested):

app/test_main.py
```diff
@@ -1 +1,53 @@
...
+                                is_internet_connection: bool,
+                                expected_result: str,
+                                url: str) -> None:
+    mocked_valid_google_url.return_value = is_valid_url
+    mocked_has_internet_connection.return_value = is_internet_connection
+
+    assert can_access_google_page(url) == expected_result
+
+
+@patch("app.main.has_internet_connection")
```
Oleksl888, line 42: if you are reusing your mocked functions in different tests it would be better design to put them in a fixture

Oleksl888 (approved):
Good job!

PR #822, opened 2024-04-03
solution

alonasorochynska:


app/test_main.py
```diff
@@ -1 +1,44 @@
...
+    assert can_access_google_page("https://www.google.com") == "Not accessible"
+
+
+def test_can_access_google_page_with_no_internet_and_valid_url(
+        mocked_url: mock.MagicMock,
+        mocked_internet: mock.MagicMock
+) -> None:
+    mocked_internet.return_value = False
+    mocked_url.return_value = True
+    assert can_access_google_page("https://www.google.com") == "Not accessible"
```
YuriiKindrat, line 44: could u use parametrize here, to optimize ur code ?
also what about false/false case ?

YuriiKindrat (changes requested):

app/test_main.py
```diff
@@ -1 +1,68 @@
-# write your code here
+from unittest import mock
+import pytest
+
+
+from app.main import can_access_google_page
+
+
+@pytest.mark.parametrize(
```
Oleksl888, line 8: You should parametrize the function that you are going to test and pass this data to.

app/test_main.py
```diff
@@ -1 +1,68 @@
...
+def mocked_internet() -> None:
+    with mock.patch("app.main.has_internet_connection") as mocked_time:
+        yield mocked_time
+
+
+def test_can_access_google_page_with_internet_and_valid_url(
+        mocked_url: mock.MagicMock,
+        mocked_internet: mock.MagicMock
+) -> None:
+    mocked_url.return_value = True
```
Oleksl888, line 39: You added parameters but you still hardcode return values here.
The correct implementation will look something like this:
1. Create one test function with assert that will check for correct result and asserts that mock function has been called.
2. Add parametrizer to that test and use values from parametrizer instead of hard coding.
3. You should mock required functions btw.

Oleksl888 (changes requested):

alonasorochynska:
I don´t know if this decision is correct but i can not understand why this code passes pytests only if one mock is located inside another one. Why it is not working if to place them one by one?

app/test_main.py
```diff
@@ -1 +1,38 @@
-# write your code here
+from unittest import mock
```
Oleksl888, line 1: must add line between different types of imports. 1 line is enough

app/test_main.py
```diff
@@ -1 +1,38 @@
...
+        mocked_url.return_value = valid_url
+
+        with mock.patch(
+                "app.main.has_internet_connection"
+        ) as mocked_connection:
+            mocked_connection.return_value = valid_internet
+
+            assert can_access_google_page(
+                "https://www.google.com"
+            ) == expected_result
```
Oleksl888, line 38: what about checking if mocked functions has been called?

Oleksl888 (changes requested):

YuriiKindrat (approved):
GJ

PR #823, opened 2024-04-04
Solution

Nyxan:


app/test_main.py
```diff
@@ -1 +1,24 @@
-# write your code here
+import pytest
+from unittest.mock import patch
```
Oleksl888, line 2: check order of imports, also different import types must be separated with blank lines

app/test_main.py
```diff
@@ -1 +1,24 @@
...
+         "Non-existent URL is internet connection",
+         "Existing URL  no internet connection"]
+)
+def test_can_access_google_page(valid_url: str,
+                                internet_connection: bool,
+                                expected_result: str) -> None:
+    with patch("app.main.has_internet_connection",
+               return_value=internet_connection):
+        with patch("app.main.valid_google_url",
+                   return_value=(valid_url == "https://www.google.com")):
```
Oleksl888, line 23: line overindented

app/test_main.py
```diff
@@ -1 +1,24 @@
...
+from unittest.mock import patch
+from app.main import can_access_google_page
+
+
+@pytest.mark.parametrize(
+    "valid_url,internet_connection,expected_result",
+    [
+        ("https://www.google.com", True, "Accessible"),
+        ("https://non-existent_google.com", True, "Not accessible"),
+        ("https://www.google.com", False, "Not accessible")
```
Oleksl888, line 11: missing case when both mock functions produce False

app/test_main.py
```diff
@@ -1 +1,24 @@
...
+         "Existing URL  no internet connection"]
+)
+def test_can_access_google_page(valid_url: str,
+                                internet_connection: bool,
+                                expected_result: str) -> None:
+    with patch("app.main.has_internet_connection",
+               return_value=internet_connection):
+        with patch("app.main.valid_google_url",
+                   return_value=(valid_url == "https://www.google.com")):
+            assert can_access_google_page(valid_url) == expected_result
```
Oleksl888, line 24: missing assert for mocked functions being or not being called

Oleksl888 (changes requested):

app/test_main.py
```diff
@@ -1 +1,39 @@
...
+        expected_result: str
+) -> None:
+    mock_has_internet_connection = MagicMock(return_value=internet_connection)
+    mock_valid_google_url = (
+        MagicMock(return_value=(valid_url == "https://www.google.com"))
+    )
+
+    with patch("app.main.has_internet_connection",
+               mock_has_internet_connection):
+        with patch("app.main.valid_google_url", mock_valid_google_url):
```
Oleksl888, line 33: Why don't you simplify this by using @patch decorators? it will make code more readable. 
also `MagicMock(return_value=(valid_url == "https://www.google.com"))` this and putting different URLs in params is an overkill. All we need to do with mocked functions is to set return value to True or False so you can simplify this as well

Oleksl888 (changes requested):

PR #824, opened 2024-04-04
Solution

SerhiiMar:


app/test_main.py
```diff
@@ -1 +1,35 @@
-# write your code here
+from typing import Callable
+from unittest.mock import patch
+
+import pytest
+
+from app.main import can_access_google_page
```
Oleksl888, line 6: I like how you structured imports 👏

app/test_main.py
```diff
@@ -1 +1,35 @@
...
+        "return 'Accessible' if url is valid and there is internet access",
+        "return 'Not accessible' if url is not valid",
+        "return 'Not accessible' if there is no internet",
+        "return 'Not accessible' if url is not valid and there is no internet",
+    ]
+)
+@patch("app.main.valid_google_url")
+@patch("app.main.has_internet_connection")
+def test_can_access_google_page(
+    mocked_internet_func: Callable,
```
Oleksl888, line 27: mock has its own distinct type

app/test_main.py
```diff
@@ -1 +1,35 @@
...
+    mocked_url_func: Callable,
+    is_internet: bool,
+    is_url: bool,
+    expected: str
+) -> None:
+    mocked_internet_func.return_value = is_internet
+    mocked_url_func.return_value = is_url
+    assert can_access_google_page("url") == expected
```
Oleksl888, line 35: shouldn't you also test that your mocked functions had or had not been called?

Oleksl888 (changes requested):

app/test_main.py
```diff
@@ -1 +1,54 @@
...
+
+
+@patch("app.main.valid_google_url", return_value=True)
+@patch("app.main.has_internet_connection", return_value=True)
+class TestInternalFuncCalled:
+    def test_has_internet_connection_called(
+        self,
+        mocked_internet_func: MagicMock,
+        mocked_url_func: MagicMock
```
Oleksl888, line 43: mocked_url is not used in this method

app/test_main.py
```diff
@@ -1 +1,54 @@
...
+    ) -> None:
+        can_access_google_page("url")
+        mocked_internet_func.assert_called_once()
+
+    def test_valid_google_url_called(
+        self,
+        mocked_internet_func: MagicMock,
+        mocked_url_func: MagicMock
+    ) -> None:
+        can_access_google_page("url")
```
Oleksl888, line 53: "url" is reused throughout different tests - so it would rather be a fixture

Oleksl888 (changes requested):

app/test_main.py
```diff
@@ -1 +1,76 @@
...
+        mocked_url_func: MagicMock,
+        is_internet: bool,
+        is_url: bool,
+        expected: int,
+        url: str,
+    ) -> None:
+        mocked_internet_func.return_value = is_internet
+        mocked_url_func.return_value = is_url
+        can_access_google_page(url)
+        assert mocked_internet_func.called + mocked_url_func.called >= expected
```
Oleksl888, line 76: this is clever but why use >= here? also shouldn't False + False result in 0?

Oleksl888 (changes requested):

app/test_main.py
SerhiiMar: I use `>=` because it's test and i don't know how many times tested code called the function.
In `True` + `True` case test checks if each function called at least once.
In other cases test checks if one of the functions called at least once because:
```
if has_internet_connection() and valid_google_url(url): ...
```

```
result1 = has_internet_connection()
result2 = valid_google_url(url)
if result1 and result2: ...
```
In `False` + `False` case result not in 0 because `func.return_value != func.called`

Oleksl888 (approved):
you got it!

PR #825, opened 2024-04-04
Solution for task can access google page

oleksandra-shershen:


app/test_main.py
```diff
@@ -1 +1,40 @@
-# write your code here
+import pytest
+from unittest.mock import patch, MagicMock
```
Oleksl888, line 2: Fix order of imports please and add blank lines between different import types

app/test_main.py
```diff
@@ -1 +1,40 @@
...
+from freezegun import freeze_time
+
+
+@pytest.fixture
+def mock_valid_google_url() -> MagicMock:
+    with patch("app.main.requests.get") as mock_get:
+        yield mock_get
+
+
+def test_accessibility_with_valid_connection(
```
Oleksl888, line 13: You made three test which do exactly the same thing hence you can make one test and parametrize it.
Also you must have cover all combinations of possible responses of both mocked functions

app/test_main.py
```diff
@@ -1 +1,40 @@
...
+        mock_valid_google_url: MagicMock
+) -> None:
+    mock_valid_google_url.return_value.status_code = 200
+
+    with freeze_time("2024-04-04 10:00:00"):
```
Oleksl888, line 18: freeze gun is an overkill here - setting expected result of mocked functions  to True/False will be enough

Oleksl888 (changes requested):

app/test_main.py
```diff
@@ -1 +1,51 @@
-# write your code here
+import pytest
+from unittest import mock
+from unittest.mock import MagicMock
+from app.main import can_access_google_page
+
+
+@pytest.fixture
```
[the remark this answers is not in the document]
oleksandra-shershen: Resolved

Oleksl888 (approved):
Nice!

PR #826, opened 2024-04-04
Solution 1

semermaks:


app/test_main.py
```diff
@@ -1 +1,56 @@
...
+    ids=[
+        "can't access only with connection",
+        "can't access only with valid url",
+        "can't access with no connection and valid url",
+        "can access with connection and valid url"
+    ]
+)
+def test_can_access_google_page(
+    url: str,
+    mock_internet_connection: Callable,
```
Oleksl888, line 48: wrong type for mock object

app/test_main.py
```diff
@@ -1 +1,56 @@
...
+    mock_valid_google_url: Callable,
+    connection_return: bool,
+    valid_url_return: bool,
+    expected_result: str
+) -> None:
+    mock_internet_connection.return_value = connection_return
+    mock_valid_google_url.return_value = valid_url_return
+    assert can_access_google_page(url) == expected_result
```
Oleksl888, line 56: what about checking if your mocked functions has been called?

app/test_main.py
```diff
@@ -1 +1,56 @@
...
+
+
+@pytest.fixture
+def url() -> str:
+    return ("https://www.google.com/webhp?hl=en&sa="
+            "X&ved=0ahUKEwjlr5aw25eFAxVFFBAIHbjlAdEQPAgJ")
+
+
+@pytest.fixture
+def mock_internet_connection(monkeypatch: MonkeyPatch) -> Callable:
```
Oleksl888, line 15: Can you please explain why did you choose going through all the trouble of making fixtures and using monkey patch instead of making mock in one line of code in your test function?

app/test_main.py
```diff
@@ -1 +1,56 @@
-# write your code here
+import pytest
+from _pytest.monkeypatch import MonkeyPatch
+from typing import Callable
+from unittest.mock import MagicMock
+from app.main import can_access_google_page
+
+
+@pytest.fixture
+def url() -> str:
```
Oleksl888, line 9: you don't really need this fixture - you use it only in one place and since your functions are mocked it doesn't really matter what url is

Oleksl888 (changes requested):

app/test_main.py
```diff
@@ -1 +1,35 @@
-# write your code here
+from unittest import mock
+import pytest
```
Oleksl888, line 2: must be blank lines between different import types

Oleksl888 (changes requested):

LiudmylaKulinchenko (approved):
Well done 🎉

PR #827, opened 2024-04-04
Solution

SashaKisliy:


app/test_main.py
```diff
@@ -1 +1,45 @@
-# write your code here
+from unittest import mock
+import pytest
```
Oleksl888, line 2: need to add blank lines between different import types

app/test_main.py
```diff
@@ -1 +1,45 @@
...
+        has_valid_url: bool,
+        expected: str
+) -> None:
+    mock_valid_url.return_value = has_valid_url
+    mock_internet_connection.return_value = has_internet
+
+    assert can_access_google_page("https://www.google.com/") == expected
+
+
+@mock.patch("app.main.has_internet_connection")
```
Oleksl888, line 35: There is a lot of repetitive code here. If you want to keep it as two separate tests then you can put mocks into a fixture and reuse it in different tests.

app/test_main.py
```diff
@@ -1 +1,45 @@
...
+@mock.patch("app.main.valid_google_url")
+def test_func_callable(
+        mock_valid_url: mock.MagicMock,
+        mock_internet_connection: mock.MagicMock
+) -> None:
+    url = "https://www.google.com/"
+    can_access_google_page(url)
```
Oleksl888, line 42: you also reuse url in both test which can also become a fixture

Oleksl888 (changes requested):

Oleksl888 (approved):
Well done!

PR #829, opened 2024-04-04
can access google

mikhailyemets:


Oleksl888 (approved):
Very nice sir 👏

app/test_main.py
```diff
@@ -1 +1,40 @@
-# write your code here
+from unittest import mock
```
Oleksl888, line 1: you could actually import `from unittest.mock import MagicMock, patch`
so you would take only what you need

PR #831, opened 2024-04-04
Solution#1

VitalyBashkiser:


app/test_main.py
```diff
@@ -1 +1,34 @@
-# write your code here
+from unittest import mock
```
Oleksl888, line 1: make space between different import types

app/test_main.py
```diff
@@ -1 +1,34 @@
...
+import pytest
+from app.main import can_access_google_page
+
+
+@pytest.mark.parametrize(
+    "url, valid_url, internet_connection, expected",
+    [
+        # Valid URL and internet connection
```
Oleksl888, line 9: you should use id/ids instead of writing comments like this

app/test_main.py
```diff
@@ -1 +1,34 @@
...
+        # Valid URL but no internet connection
+        ("https://www.google.com", True, False, "Not accessible"),
+        # Invalid URL and no internet connection
+        ("https://www.example.com", False, False, "Not accessible"),
+    ]
+)
+@mock.patch("app.main.valid_google_url")
+@mock.patch("app.main.has_internet_connection")
+def test_can_access_google_page(
+    mock_internet_connection: mock.Mock,
```
Oleksl888, line 22: it is actually of type MagicMock to be specific

app/test_main.py
```diff
@@ -1 +1,34 @@
...
+    valid_url: bool,
+    internet_connection: bool,
+    expected: str
+) -> None:
+    # Setting up the mock return values
+    mock_valid_url.return_value = valid_url
+    mock_internet_connection.return_value = internet_connection
+
+    # Calling the function with the provided URL
+    assert can_access_google_page(url) == expected
```
Oleksl888, line 34: what about checking if your mocked functions has been called?

app/test_main.py
```diff
@@ -1 +1,34 @@
...
+import pytest
+from app.main import can_access_google_page
+
+
+@pytest.mark.parametrize(
+    "url, valid_url, internet_connection, expected",
+    [
+        # Valid URL and internet connection
+        ("https://www.google.com", True, True, "Accessible"),
+        # Invalid URL but internet connection
```
Oleksl888, line 11: don't think you need to parametrize url - it can be anything since you mock function that actually does a request

app/test_main.py
```diff
@@ -1 +1,34 @@
...
+@mock.patch("app.main.has_internet_connection")
+def test_can_access_google_page(
+    mock_internet_connection: mock.Mock,
+    mock_valid_url: mock.Mock,
+    url: str,
+    valid_url: bool,
+    internet_connection: bool,
+    expected: str
+) -> None:
+    # Setting up the mock return values
```
Oleksl888, line 29: remove self explanatory comments

Oleksl888 (changes requested):

app/test_main.py
```diff
@@ -1 +1,42 @@
...
+    mock_has_internet_connection.return_value = internet_connection
+
+    can_access_google_page("https://anylink.com")
+
+    mock_has_internet_connection.assert_called_once()
+
+    if mock_has_internet_connection.assert_called_once():
+        mock_valid_url.assert_called_once_with("https://anylink.com")
+
+    assert can_access_google_page("https://anylink.com") == expected
```
Oleksl888, line 42: no need to call `can_access_google_page` twice.
also since you reuse `"https://anylink.com"` it must be stored in a variable

Oleksl888 (changes requested):

YuriiKindrat (approved):
GJ

PR #832, opened 2024-04-04
Implement tests for Can access Google page task

bwaah1:


app/test_main.py
```diff
@@ -1 +1,41 @@
-# write your code here
+import pytest
+from unittest import mock
```
Oleksl888, line 2: check order of imports - also optimize them. No need to import mock on this line and then import from mock on line 4

app/test_main.py
```diff
@@ -1 +1,41 @@
...
+    with (mock.patch("app.main.has_internet_connection")
+          as mocked_has_internet_connection):
+        yield mocked_has_internet_connection
+
+
+def test_can_access_google_page(
+        mock_valid_google_url: MagicMock,
+        mock_has_internet_connection: MagicMock
+) -> None:
+    mock_valid_google_url.return_value = True
```
Oleksl888, line 25: There is a lot of repetition here - you must parametrize this test

app/test_main.py
```diff
@@ -1 +1,41 @@
...
+    mock_valid_google_url.return_value = False
+    mock_has_internet_connection.return_value = True
+
+    assert (can_access_google_page("https://www.[sdfl;dsf].com/")
+            == "Not accessible")
+
+    mock_valid_google_url.return_value = True
+    mock_has_internet_connection.return_value = False
+
+    assert (can_access_google_page("https://www.google.com/")
```
Oleksl888, line 40: don't  forget to check if your mocked functions has been called

app/test_main.py
```diff
@@ -1 +1,41 @@
...
+
+@pytest.fixture
+def mock_valid_google_url() -> MagicMock:
+    with (mock.patch("app.main.valid_google_url")
+          as mocked_valid_google_url):
+        yield mocked_valid_google_url
+
+
+@pytest.fixture
+def mock_has_internet_connection() -> MagicMock:
```
Oleksl888, line 15: it would make sense to put mocks in a fixture if you have many test and you are going to reuse them. but since you have only one test why you use fixtures here?

app/test_main.py
```diff
@@ -1 +1,41 @@
-# write your code here
+import pytest
+from unittest import mock
+from app.main import can_access_google_page
+from mock import MagicMock
+
+
```
Oleksl888, line 6: Also this code doesnt pass the tests

Oleksl888 (changes requested):

LiudmylaKulinchenko (approved):
Good job 🌷

PR #835, opened 2024-04-07
success

Olevsky-Ivan:


tests/test_tests.py
Dimosphen1, line 1: Remove changes done to this file, it should be left untouched

app/test_main.py
```diff
@@ -1 +1,25 @@
...
+            with patch(
+                    "app.main.valid_google_url",
+                    return_value=True
+            ):
+                with patch(
+                        "app.main.can_access_google_page",
+                        return_value=True
+                ):
+                    response = requests.get(self.url)
+                    self.assertEqual(response.status_code, 200)
```
Dimosphen1, line 25: You should test 4 cases, when `has_internet_connection` and `valid_google_url` return either True/True, False/True, True/False or False/False. You could use `@pytest.mark.parametrize` to avoid code duplication as use mocks as decorators

Dimosphen1 (changes requested):

app/test_main.py
Olevsky-Ivan: Can do it so?

PR #837, opened 2024-04-07
'Solution1'

yossosh:


yossosh:
I do not know what to do to solve this problem :( 
please tell me how to solve it

Oleksl888:
> I do not know what to do to solve this problem :( please tell me how to solve it

the tests fail here

> def test_can_access_google_page(mocker: MagicMock) -> None:
E       fixture 'mocker' not found

It looks like your test is expecting a `mocker` parameter which doesn't exist. Your code should look something like this:
1. You should use @parametrize decorator on your test to pass the testing parameters.
2. You should use @patch on your test to mock the necessary functions.
3. Your test function will take values produced by above decorators as inputs and you use them in your assert statements

Oleksl888 (changes requested):
Suggesting to attend  QnA if you didn't figure out this yet

MNDonut (approved):

PR #839, opened 2024-04-07
Solution

siaserb:


app/test_main.py
```diff
@@ -1 +1,60 @@
...
+
+
+@pytest.fixture
+def mocked_connection() -> Any:
+    with mock.patch("app.main.has_internet_connection") as mocked_connection:
+        yield mocked_connection
+
+
+def test_can_access_google_page_when_validation_and_connection(
+        mocked_validation: Any,
```
Oleksl888, line 21: You can write one test instead of four and parametrize inputs since your test do the same thing but with different params.
Also don't forget to test that your mocked functions has or has not been called

Oleksl888 (changes requested):

app/test_main.py
```diff
@@ -1 +1,60 @@
-# write your code here
+import pytest
+from unittest import mock
```
Oleksl888, line 2: check order of imports  + add blank lines between

app/test_main.py
```diff
@@ -1 +1,60 @@
...
+from typing import Any
+
+from app.main import can_access_google_page
+
+
+@pytest.fixture
+def mocked_validation() -> Any:
```
Oleksl888, line 9: mock has its own type

app/test_main.py
```diff
@@ -1 +1,60 @@
...
+
+        pytest.param(False,
+                     False,
+                     "Not accessible",
+                     id="when no validation and no connection "
+                        "are true, page should not be accessible")
+    ]
+)
+def test_can_access_google_page(
+        mocked_validation: Any,
```
Oleksl888, line 49: please put correct typing for mocked functions

app/test_main.py
```diff
@@ -1 +1,60 @@
...
+        mocked_connection: Any,
+        validation_return: bool,
+        connection_return: bool,
+        access_return: str
+) -> None:
+    can_access_google_page("")
+    mocked_validation.assert_called_once()
+    mocked_connection.assert_called_once()
+    mocked_validation.return_value = validation_return
```
Oleksl888, line 58: here the order is a bit broken - first we need to set return values for mocks, then call main function, then check if mocked functions has been called

Oleksl888 (changes requested):

app/test_main.py
```diff
@@ -1 +1,62 @@
-# write your code here
+import pytest
+
+from unittest import mock
+
+from app.main import can_access_google_page
+
+
+@pytest.fixture
```
Oleksl888, line 8: you have only one test here so it will be more appropriate to add decorators on top of the function
`@mock.patch("app.main.valid_google_url")`

app/test_main.py
```diff
@@ -1 +1,62 @@
...
+def mocked_validation() -> mock.Mock:
```
Oleksl888, line 9: Also this mock will be of MagicMock type to be specific

app/test_main.py
```diff
@@ -1 +1,62 @@
...
+        mocked_connection: mock.Mock,
+        validation_return: bool,
+        connection_return: bool,
+        access_return: str
+) -> None:
+
+    mocked_validation.return_value = validation_return
+    mocked_connection.return_value = connection_return
+    assert can_access_google_page("") == access_return
+    if mocked_connection():
```
Oleksl888, line 59: here `has_internet_connection` will be called always and `valid_google_url` will be called only if `has_internet_connection` returned True

Oleksl888 (approved):
Approved, but please check the comments and improve

PR #840, opened 2024-04-07
Solution

VladyslavKsh:


app/test_main.py
```diff
@@ -1 +1,39 @@
-# write your code here
+from unittest import mock
+
+
+from app.main import can_access_google_page
+
+
+class TestAccessPage:
+    def test_access_google_page_with_invalid_url(self) -> None:
+        with mock.patch("app.main.valid_google_url")\
```
Oleksl888, line 9: We don't use '\' for breaking lines in our projects

app/test_main.py
```diff
@@ -1 +1,39 @@
...
+             as mock_valid_google_url,\
+             mock.patch("app.main.has_internet_connection")\
```
Oleksl888, line 11: you have three tests that do the same thing and the only difference is return values of mocked functions. Would make more sense to use parametrize here

app/test_main.py
```diff
@@ -1 +1,39 @@
...
+                   "Not accessible"
+
+    def test_access_google_page_with_url_and_internet(self) -> None:
+        with mock.patch("app.main.valid_google_url")\
+             as mock_valid_google_url,\
+             mock.patch("app.main.has_internet_connection")\
+             as mock_has_internet_connection:
+
+            mock_valid_google_url.return_value = True
+            mock_has_internet_connection.return_value = True
```
Oleksl888, line 37: You also missing a case when both mock functions produce False result and test that the mocked functions has or has not been called

app/test_main.py
```diff
@@ -1 +1,39 @@
...
+             mock.patch("app.main.has_internet_connection")\
+             as mock_has_internet_connection:
+
+            mock_valid_google_url.return_value = False
+            mock_has_internet_connection.return_value = True
+            assert can_access_google_page("https://example.com") ==\
+                   "Not accessible"
+
+    def test_access_google_page_without_internet(self) -> None:
+        with mock.patch("app.main.valid_google_url")\
```
Oleksl888, line 20: using decorators for patching may make your code more readable

Oleksl888 (changes requested):

app/test_main.py
```diff
@@ -1 +1,71 @@
...
+    mock_has_internet_connection: mock.MagicMock,
+    mock_valid_google_url: mock.MagicMock,
+    valid_google_url_value: bool,
+    has_internet_connection_value: bool,
+    expectation: str
+) -> None:
+    mock_valid_google_url.return_value = valid_google_url_value
+    mock_has_internet_connection.return_value = has_internet_connection_value
+    assert app.main.can_access_google_page("https://www.google.com")\
+           == expectation
```
YuriiKindrat, line 49: could u break the line with brackets here ?

YuriiKindrat (changes requested):

app/test_main.py
```diff
@@ -1 +1,73 @@
...
+) -> None:
+    mock_valid_google_url.return_value = valid_google_url_value
+    mock_has_internet_connection.return_value = has_internet_connection_value
+    assert (
+        app.main.can_access_google_page("https://www.google.com")
+        == expectation
+    )
+
+
+@mock.patch("app.main.valid_google_url")
```
Oleksl888, line 54: would make sense to test these functions with different param combinations - you will get different results if `has_internet_connection` function returns True or False

Oleksl888: This solution is not optimal - you repeat the same test 4 times, instead of making one test with 4 different sets of params - same as `test_access_google_page`

app/test_main.py
```diff
@@ -1 +1,106 @@
-# write your code here
+import pytest
+
+import app.main
+
+from unittest import mock
```
Oleksl888, line 5: this import must be on top because it is built-in.

Oleksl888 (approved):
Approved, but please look through the comments and make it right

PR #842, opened 2024-04-08
Solution

Ruxxell54:


app/test_main.py
```diff
@@ -1 +1,38 @@
...
+
+
+class TestCanAccessGooglePage(unittest.TestCase):
+
+    @patch("app.main.has_internet_connection")
+    @patch("app.main.valid_google_url")
+    def test_can_access_google_page_accessible(
+            self,
+            mock_valid_google_url: MagicMock,
+            mock_has_internet_connection: MagicMock) -> None:
```
YuriiKindrat, line 13: bad formatting here (

app/test_main.py
```diff
@@ -1 +1,38 @@
...
+        result = can_access_google_page("http://www.google.com")
+
+        self.assertEqual(result, "Accessible")
+
+    @patch("app.main.has_internet_connection")
+    @patch("app.main.valid_google_url")
+    def test_can_access_google_page_not_accessible(
+            self,
+            mock_valid_google_url: MagicMock,
+            mock_has_internet_connection: MagicMock) -> None:
```
YuriiKindrat, line 27: same

YuriiKindrat (changes requested):
use parametrize here to test all cases 
e.g. True,True/False,True/True,False/False,False

PR #845, opened 2024-04-08
Solution

Nyxan:


app/test_main.py
```diff
@@ -1 +1,40 @@
...
+
+
+@pytest.mark.parametrize(
+    "valid_url,internet_connection,expected_result",
+    [
+        ("https://www.google.com", True, "Accessible"),
+        ("https://non-existent_google.com", True, "Not accessible"),
+        ("https://www.google.com", False, "Not accessible"),
+        ("https://non-existent_google.com", False, "Not accessible")
+    ],
```
YuriiKindrat, line 15: no need to use urls here, u can mock the return value (True, False) of a func

app/test_main.py
```diff
@@ -1 +1,40 @@
...
+def test_can_access_google_page(
+    mock_valid_google_url: MagicMock,
+    mock_has_internet_connection: MagicMock,
+    valid_url: str,
+    internet_connection: bool,
+    expected_result: str
+) -> None:
+    mock_has_internet_connection.return_value = internet_connection
+    mock_valid_google_url.return_value = \
+        (valid_url == "https://www.google.com")
```
YuriiKindrat, line 32: try to avoid using linebreakers in ur code

YuriiKindrat (changes requested):

YuriiKindrat (approved):
GJ

PR #849, opened 2024-04-10
Develop

RomanDanyl:


app/test_main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+from unittest import mock
+from app.main import can_access_google_page
+
+
+def test_has_internet_connection_was_called() -> None:
+    with mock.patch(
+            "app.main.has_internet_connection") as mocked_has_connection, \
+            mock.patch("app.main.valid_google_url") as mocked_valid_url:
```
EdAlekseiev, line 8: Use `()` instead of `\` for separating lines

EdAlekseiev (changes requested):
You should add test cases for testing result of can_access_google_page

app/test_main.py
```diff
@@ -1 +1,39 @@
-# write your code here
+import pytest
+from unittest import mock
+from app.main import can_access_google_page
```
Dimosphen1, line 3: Order and separate your imports correctly to follow PEP-8

app/test_main.py
```diff
@@ -1 +1,39 @@
...
+    with mock.patch(
+        "app.main.has_internet_connection"
+    ) as mocked_has_connection, mock.patch(
+        "app.main.valid_google_url"
+    ) as mocked_valid_url:
+
+        url = "https://www.google.com"
+        can_access_google_page(url)
+        mocked_has_connection.assert_called_once()
+        mocked_valid_url.assert_called_once()
```
Dimosphen1, line 39: Please, create one function that will serve all checks. You should mock the return value for `valid_google_url` and `has_internet_connection`, not their calls or hours. Your input to `@pytest.mark.parametrize` should be something like `"valid_url,has_connection,expected"` and you should test 4 cases: when `valid_google_url` and `has_internet_connection` return either True/True, False/True, True/False, False/False. Besides that, it is preferable to use mocks as decorators to not initialize context manager for a whole function body

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #853, opened 2024-04-11
First_try

Sydorovich:


NazarSheremet (changes requested):
In this task will be better to use `@pytest.mark.parametrize` It will simpify your code

PR #857, opened 2024-04-14
Solution

frontbastard:


frontbastard:
I feel that there is a much more easier solution for this task. If so, could you show?

app/test_main.py
```diff
@@ -1 +1,38 @@
...
+
+from app.main import can_access_google_page
+
+
+@pytest.mark.parametrize(
+    "valid_url, internet_available, expected_result, url_called",
+    [
+        (True, True, "Accessible", True),
+        (False, True, "Not accessible", True),
+        (True, False, "Not accessible", False)
```
vsmutok, line 13: What about False and False?

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #873, opened 2024-04-28
SOL

neagarr:


app/test_main.py
```diff
@@ -1 +1,25 @@
...
+import pytest
+from unittest import mock
+from app.main import can_access_google_page
+
+
+@pytest.mark.parametrize(
+    "connection,url,result",
+    [
+        (True, False, "Not accessible"),
+        (False, True, "Not accessible")
```
vsmutok, line 10: What about other case?
`True, True` and `False, False`

vsmutok (changes requested):

PR #884, opened 2024-05-10
triukhan solution

triukhan:


app/test_main.py
```diff
@@ -1 +1,23 @@
...
+from app.main import can_access_google_page
+from typing import Callable
+import pytest
+
+
+@pytest.mark.parametrize(
+    "values,result", [
+        ([True, True], "Accessible"),
+        ([True, False], "Not accessible"),
+        ([False, True], "Not accessible"),
```
LiudmylaKulinchenko, line 11: Add error messages. It's good when the user understands what the problem is

app/test_main.py
```diff
@@ -1 +1,23 @@
...
+    ]
+)
+@mock.patch("app.main.valid_google_url")
+@mock.patch("app.main.has_internet_connection")
```
LiudmylaKulinchenko, line 15: U should add a test for False, False

LiudmylaKulinchenko (changes requested):

app/test_main.py
```diff
@@ -1 +1,28 @@
...
+from unittest import mock
+from app.main import can_access_google_page
+from typing import Callable
+import pytest
+
+
+@pytest.mark.parametrize(
+    "values,result", [
+        pytest.param([True, True], "Accessible",
+                     id="User don't have access with internet and valid url"),
```
LiudmylaKulinchenko, line 10: Why doesn't use has access with internet and valid url?)

app/test_main.py
```diff
@@ -1 +1,28 @@
...
+        pytest.param([True, False], "Not accessible",
+                     id="User has access without valid url"),
+        pytest.param([False, True], "Not accessible",
+                     id="User has access without internet"),
+        pytest.param([False, False], "Not accessible",
+                     id="User has access without internet and valid url"),
```
LiudmylaKulinchenko, line 16: Fix all your messages

LiudmylaKulinchenko: `returns "Accessible" if url is valid to access the Google home page and it has internet connection, else it returns "Not accessible".`

LiudmylaKulinchenko (changes requested):

PR #886, opened 2024-05-13
Develop

include031:


vasylhnatiuk (changes requested):
you should use @pytest.mark.parametrize and write test with different values

PR #890, opened 2024-05-20
GooglePageSolved

YakovK1808:
At my first implementation, test_main wasn't able to locate the main.py module, I tried to add the line "from app.main" instead of "from main" and a few other things and it didn't work. I asked GPT for help and it said that I should use:
import sys
import os

sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))

and it worked, I hope it's fine.

### DIFF
--- a/app/test_main.py
+++ b/app/test_main.py
@@ -1,1 +1,43 @@
-# write your code here
+import unittest
+from unittest.mock import patch
+import sys
+import os
+from app.main import can_access_google_page
+from typing import Any
+
+sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+
+
+class TestCanAccessGooglePage(unittest.TestCase):
+
+    @patch("app.main.valid_google_url")
+    @patch("app.main.has_internet_connection")
+    def test_accessible_when_both_conditions_true(self, mock_has_internet_connection: Any, mock_valid_google_url: Any) -> None:  # noqa: E501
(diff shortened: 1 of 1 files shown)


app/test_main.py
```diff
@@ -1 +1,43 @@
...
+    @patch("app.main.valid_google_url")
+    @patch("app.main.has_internet_connection")
+    def test_accessible_when_both_conditions_true(self, mock_has_internet_connection: Any, mock_valid_google_url: Any) -> None:  # noqa: E501
+        mock_has_internet_connection.return_value = True
+        mock_valid_google_url.return_value = True
+        self.assertEqual(can_access_google_page("http://www.google.com"), "Accessible")  # noqa: E501
+
+    @patch("app.main.valid_google_url")
+    @patch("app.main.has_internet_connection")
+    def test_not_accessible_when_no_internet_connection(self, mock_has_internet_connection: Any, mock_valid_google_url: Any) -> None:  # noqa: E501
```
NazarSheremet, line 22: You can do like this to not mute flake
```suggestion
    def test_not_accessible_when_no_internet_connection(
       self,
       mock_has_internet_connection: Any,
       mock_valid_google_url: Any) -> None:
```

NazarSheremet (changes requested):
Will be better to use `@pytest.mark.parametrize` to not dublicate the same code

YuriiSmolii (approved):
LGTM

PR #893, opened 2024-05-23
Hell yeah solution

RuslanLevchuk:


vasylhnatiuk (changes requested):
add tests for         mocked_internet.return_value = False, mocked_url_validation.return_value = False, and True, True. And it would be better to use pytest.parametrize

RuslanLevchuk:
ok, thanx

vasylhnatiuk (approved):

PR #894, opened 2024-05-23
added tests to can access google page

AlexRozv:
solution


LiudmylaKulinchenko (changes requested):
You can also consider using parametrize here with descriptive messages, since test logic is similar

LiudmylaKulinchenko (approved):
Good job 🔥

PR #900, opened 2024-05-28
solution

AndrewwSSS:


app/test_main.py
```diff
@@ -1 +1,25 @@
...
+from app.main import can_access_google_page
+
+
+@pytest.mark.parametrize(
+    "url, mock_valid_url, mock_internet, expected",
+    [
+        ("https://www.google.com", True, True, "Accessible"),
+        ("https://www.google.com", True, False, "Not accessible"),
+        ("https://invalid-url.com", False, True, "Not accessible"),
+        ("https://invalid-url.com", False, False, "Not accessible"),
```
Oleg-s-s, line 12: Add descriptive error messages. It's good when the user understands what the problem is.

Oleg-s-s (changes requested):

DevSpaciX (approved):

PR #901, opened 2024-05-28
Solution

SuskyiVolodymyr:


app/test_main.py
```diff
@@ -1 +1,51 @@
...
+        ),
+        pytest.param(
+            "https://www.google.com",
+            "Accessible",
+            True,
+            True,
+            id="Accessible when has internet and valid url"
+        ),
+    ]
+)
```
Oleg-s-s, line 38: Its all good but you can shorten it using this type of approach:
`@pytest.mark.parametrize("name",[value],ids=[id])`

Oleg-s-s (approved):

PR #909, opened 2024-06-05
Solution

necobra:


app/test_main.py
```diff
@@ -1 +1,25 @@
...
+import pytest
+
+
+import app.main
+
+
+def test_should_access_google_page(monkeypatch: Any) -> None:
+    monkeypatch.setattr(app.main, "has_internet_connection", lambda: True)
+    monkeypatch.setattr(app.main, "valid_google_url", lambda url: True)
+    assert app.main.can_access_google_page("google.com") == "Accessible"
```
DevSpaciX, line 11: Is this test almost same as `test_should_not_access_google_page` ? You just can parametrize param `Accessible` or `Not accessible`

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #911, opened 2024-06-06
Solution

Yuruch:


app/test_main.py
```diff
@@ -1 +1,48 @@
...
+            False,
+            True,
+            "Not accessible",
+            id="Test invalid url"
+        ),
+        pytest.param(
+            False,
+            True,
+            "Not accessible",
+            id="Test valid url and have internet"
```
Oleg-s-s, line 33: What?? :eyes:

Oleg-s-s (changes requested):

Oleg-s-s (changes requested):
.

Oleg-s-s (changes requested):
.

app/test_main.py
Yuruch: forgot to change this one(

app/test_main.py
```diff
@@ -27,9 +27,9 @@
             id="Test invalid url"
         ),
         pytest.param(
-            False,
             True,
-            "Not accessible",
+            True,
+            "Accessible",
             id="Test valid url and have internet"
```
DevSpaciX, line 33: This message should describe what you will receive if the test case fails. Pls fix that in all places

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #914, opened 2024-06-07
'Solution'

dsokolovskiy:


app/test_main.py
```diff
@@ -1 +1,25 @@
...
+def test_can_access_google_page(
+        result: str,
+        google_url: bool,
+        internet_connection: bool
+) -> None:
+    with mock.patch("app.main.valid_google_url", return_value=google_url):
+        with mock.patch(
+                "app.main.has_internet_connection",
+                return_value=internet_connection
+        ):
```
DevSpaciX, line 24: Make sure your code is not very nested . Use more beautiful variant for example decoarator or something else which is less nested then now

DevSpaciX (changes requested):

app/test_main.py
dsokolovskiy: sry :(
 i did a new solution

DevSpaciX (approved):

PR #919, opened 2024-06-07
Solution_V1

BileichukIvan:


app/test_main.py
```diff
@@ -1 +1,31 @@
...
+    ]
+)
+def test_can_access_google_page(
+        mock_internet: None,
+        mock_valid_url: None,
+        expected_result: bool
+) -> None:
+    with patch("app.main.valid_google_url", return_value=mock_valid_url):
+        with patch(
+                "app.main.has_internet_connection", return_value=mock_internet
```
DevSpaciX, line 27: Always try to make your code less nested . You can patch 3 func in 1 context manager OR using decorators : )

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #938, opened 2024-06-15
Solution

MaksymProtsak:


vasylhnatiuk (changes requested):
add tests with different conditionals - not only True, True

app/test_main.py
```diff
@@ -1 +1,29 @@
...
+import pytest
+
+from app.main import can_access_google_page
+
+
+@pytest.mark.parametrize(
+    "has_net_connection, valid_url",
+    [
+        (True, True),
+        (6, True),
```
vasylhnatiuk, line 12: why do you use numbers? use different mocks like False, True or False, False

MaksymProtsak: Okey, than I should cereate one more test with False conditions and catch an errors?

app/test_main.py
```diff
@@ -1 +1,45 @@
...
+def test_can_not_access_google_page(mock_has_internet_connection: mock,
+                                    mock_valid_google_url: mock,
+                                    has_net_connection: bool | int,
+                                    valid_url: bool | int) -> None:
+    with pytest.raises(AssertionError):
+        mock_has_internet_connection.return_value = has_net_connection
+        mock_valid_google_url.return_value = valid_url
+        can_access_google_page("https://www.google.com")
+        mock_has_internet_connection.assert_called_once()
+        mock_valid_google_url.assert_called_once()
```
vsmutok, line 45: combine this into one test using `pytest.mark.parametrize`

vsmutok (changes requested):

app/test_main.py
MaksymProtsak: Ready for a check.

app/test_main.py
```diff
@@ -1 +1,35 @@
...
+
+from app.main import can_access_google_page
+
+
+@pytest.mark.parametrize(
+    "has_net_connection, valid_url, error",
+    [
+        (True, True, None),
+        (False, True, AssertionError),
+        (False, False, AssertionError)
```
vsmutok, line 13: add test for True, False case

app/test_main.py
```diff
@@ -1 +1,35 @@
...
+        can_access_google_page("https://www.google.com")
+        mock_has_internet_connection.assert_called_once()
+        mock_valid_google_url.assert_called_once()
+    except Exception:
+        with pytest.raises(error):
+            mock_has_internet_connection.return_value = has_net_connection
+            mock_valid_google_url.return_value = valid_url
+            can_access_google_page("https://www.google.com")
+            mock_has_internet_connection.assert_called_once()
+            mock_valid_google_url.assert_called_once()
```
vsmutok, line 35: no need to use `try... except` here

app/test_main.py
MaksymProtsak: Done.

app/test_main.py
MaksymProtsak: I do non anderstand, how to create separation between right and error condition without try...excep...

vasylhnatiuk (changes requested):
you don't need  try-except, use  assert like     `assert can_access_google_page("https://github.com") == final_result` to check if result is correct

vasylhnatiuk (approved):

MaksymProtsak:
🕺

PR #942, opened 2024-06-19
all tests pass

aniashev:


vasylhnatiuk (changes requested):
Add test for False, False as well

PR #943, opened 2024-06-19
solution

AndreiVed:


vasylhnatiuk (changes requested):
add test for False, False as well

vasylhnatiuk (approved):

PR #951, opened 2024-06-26
'Solution'

danylominkov:


app/test_main.py
```diff
@@ -1 +1,18 @@
...
+
+
+@mock.patch("app.main.valid_google_url")
+@mock.patch("app.main.has_internet_connection")
+def test_can_access_google_page(
+        mocked_url: bool,
+        mocked_internet_connection: bool
+) -> None:
+    mocked_url.return_value(True)
+    mocked_internet_connection.return_value(True)
```
vsmutok, line 12: as for other cases, for example, when there is no internet connection, or the url is not valid?

vsmutok (changes requested):

EdAlekseiev (changes requested):
Fix tests, please

PR #968, opened 2024-07-10
Solution

Wallian169:


app/test_main.py
```diff
@@ -1 +1,49 @@
...
+            self,
+            mock_has_internet: bool,
+            mock_valid_url: bool
+    ) -> None:
+        mock_has_internet.return_value = True
+        mock_valid_url.return_value = False
+        self.assertEqual(
+            can_access_google_page("https://invalid.url"),
+            "Not accessible"
+        )
```
Dimosphen1, line 49: - Use `parametrize` to avoid code duplication
- Test `False/False` case well for return values

Dimosphen1 (changes requested):

EdAlekseiev (approved):

PR #975, opened 2024-07-17
created tests for 'can_access_google_page' function

MasakDirt:


app/test_main.py
```diff
@@ -1 +1,41 @@
...
+from typing import Any
+from unittest import mock
+
+import pytest
+
+from app.main import can_access_google_page
+
+
+@mock.patch("app.main.has_internet_connection")
+@mock.patch("app.main.valid_google_url")
```
oliasymak24, line 10: Imagine having a fixture and not writing those decorators again and again..

app/test_main.py
```diff
@@ -1 +1,41 @@
...
+        ),
+
+        pytest.param(
+            False, True, "new url", "Not accessible",
+            id="return 'not accessible' because of internet connection"
+        ),
+
+        pytest.param(
+            True, False, "invalid@", "Not accessible",
+            id="return 'not accessible' because of invalid google url"
```
oliasymak24, line 26: U should add a test for False, False

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #980, opened 2024-07-24
Implemented solution

dimak20:


app/test_main.py
```diff
@@ -1 +1,41 @@
...
+def test_should_return_correct_string(
+        mocked_valid_google_url: Callable,
+        mocked_has_internet_connection: Callable
+) -> None:
+    assert can_access_google_page("https://www.google.com") == "Accessible"
+    mocked_valid_google_url.return_value = False
+    assert can_access_google_page("https://www.google.com") == "Not accessible"
+    mocked_valid_google_url.return_value = True
+    mocked_has_internet_connection.return_value = False
+    assert can_access_google_page("https://www.google.com") == "Not accessible"
```
oliasymak24, line 41: No need to write url for every case, u could make a constant from it

app/test_main.py
```diff
@@ -1 +1,41 @@
-# write your code here
+from unittest import mock
+from typing import Callable
+import pytest
+from app.main import can_access_google_page
+
+
```
oliasymak24, line 6: Add an error message. It's good when the user understands what the problem is

oliasymak24 (changes requested):

app/test_main.py
```diff
@@ -1 +1,57 @@
...
+from typing import Callable
+from unittest import mock
+
+import pytest
+
+from app.main import can_access_google_page
+
+GOOGLE_URL = "https://www.google.com"
+
+
```
oliasymak24, line 10: Can you try to use fot this task Use @pytest.parametrize
also there should be tests for all cases
True,true
True,False
False,True
False, False

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #981, opened 2024-07-25
Develop

danilsiv:


vasylhnatiuk (changes requested):
add also tests for False, False

danilsiv:
I removed them because the tests worked without them))

vasylhnatiuk (approved):

PR #983, opened 2024-07-25
Solution

OleksiukStepan:


app/test_main.py
```diff
@@ -1 +1,64 @@
-# write your code here
+from unittest import mock
+import pytest
+
+from app.main import can_access_google_page
+
+
+@pytest.fixture()
+def url() -> None:
```
oliasymak24, line 8: I'm not sure that this function return None?
And I'm not sure that you need a function for this

oliasymak24 (changes requested):

app/test_main.py
OleksiukStepan: Isn't it better to change the URL in one place rather than in each test?

oliasymak24 (approved):

PR #990, opened 2024-07-31
Solution

DSahalatyi:


app/test_main.py
```diff
@@ -1 +1,63 @@
...
+    )
+
+
+def test_not_accessible_without_valid_url_and_connection(
+        patch_inner_functions: Callable
+) -> None:
+    mock_url, mock_connection = patch_inner_functions
+
+    mock_url.return_value = False
+    mock_connection.return_value = False
```
oliasymak24, line 59: Use @pytest.parametrize

app/test_main.py
```diff
@@ -1 +1,63 @@
...
+
+    assert (
+        can_access_google_page("https://www.google.com/") == "Not accessible"
```
oliasymak24, line 62: No need to write url for every case, u could make a constant from it

app/test_main.py
```diff
@@ -1 +1,63 @@
-# write your code here
+from typing import Callable
+from unittest import mock
+
+import pytest
+
+from app.main import can_access_google_page
+
```
oliasymak24, line 7: Add an error message. It's good when the user understands what the problem is

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #993, opened 2024-08-05
solution

BastovOleksandr:


app/test_main.py
```diff
@@ -1 +1,46 @@
...
+    mock_valid_url.return_value = False
+
+    access_status = main.can_access_google_page("magic URL")
+
+    mock_valid_url.assert_called_once_with("magic URL")
+    assert access_status == "Not accessible"
+
+
+@mock.patch("app.main.valid_google_url")
+@mock.patch("app.main.has_internet_connection")
```
oliasymak24, line 35: Imagine having a fixture and not writing those decorators again and again..

app/test_main.py
```diff
@@ -1 +1,46 @@
...
+    mock_inet_connection: Mock,
+    mock_valid_url: Mock
+) -> None:
+    mock_inet_connection.return_value = True
+    mock_valid_url.return_value = True
+
+    access_status = main.can_access_google_page("magic URL")
+
+    mock_valid_url.assert_called_once_with("magic URL")
+    assert access_status == "Accessible"
```
oliasymak24, line 46: There should be tests for all cases
True,true
True,False
False,True
False, False

app/test_main.py
```diff
@@ -1 +1,46 @@
-# write your code here
+from unittest import mock
+from unittest.mock import Mock
+
+from app import main
+
```
oliasymak24, line 5: Add an error messages. It's good when the user understands what the problem is

app/test_main.py
```diff
@@ -1 +1,46 @@
...
+def test_should_return_accessible_if_both_inner_funcs_returns_true(
+    mock_inet_connection: Mock,
+    mock_valid_url: Mock
+) -> None:
+    mock_inet_connection.return_value = True
+    mock_valid_url.return_value = True
+
+    access_status = main.can_access_google_page("magic URL")
+
+    mock_valid_url.assert_called_once_with("magic URL")
```
oliasymak24, line 45: No need to write url for every case, u could make a constant from it

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #994, opened 2024-08-06
solution

HalynaRohatska:


app/test_main.py
```diff
@@ -1 +1,39 @@
...
+
+@pytest.mark.parametrize(
+    "has_connection, valid_url, value",
+    [
+        (True, True, "Accessible"),
+        (False, True, "Not accessible"),
+        (True, False, "Not accessible"),
+        (False, False, "Not accessible")
+    ]
+)
```
oliasymak24, line 28: Add an error messages. It's good when the user understands what the problem is

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #995, opened 2024-08-06
Solution

dartomOOv:


app/test_main.py
```diff
@@ -1 +1,53 @@
...
+
+@mark.parametrize(
+    "google_response,connection_response,expected_result",
+    [
+        (False, False, "Not accessible"),
+        (False, True, "Not accessible"),
+        (True, False, "Not accessible"),
+        (True, True, "Accessible")
+    ]
+)
```
oliasymak24, line 29: Add an error messages. It's good when the user understands what the problem is

app/test_main.py
```diff
@@ -1 +1,53 @@
...
+def test_check_if_url_fuctions_is_called(mocked_url: mock) -> None:
+    can_access_google_page("https://www.google.com")
+    mocked_url.assert_called_once()
+
+
+def test_check_if_connection_fuctions_is_called(
+        mocked_connection: mock
+) -> None:
+
+    can_access_google_page("https://www.google.com")
```
oliasymak24, line 52: No need to write url for every case, u could make a constant from it

oliasymak24 (changes requested):

PR #997, opened 2024-08-07
Solution

fposlushnyi:


app/test_main.py
```diff
@@ -1 +1,86 @@
...
+    )
+
+    def mock_has_internet_connection() -> bool:
+        return True
+    monkeypatch.setattr(
+        app.main,
+        "has_internet_connection",
+        mock_has_internet_connection
+    )
+    assert can_access_google_page(url1) == "Not accessible"
```
Dimosphen1, line 86: Avoid using `monkeypatch`, use `pytest.mark.parametrize` to avoid code duplication and `@mock.patch` decorator in solution.

app/test_main.py
```diff
@@ -1 +1,86 @@
-# write your code here
+import app.main
```
Dimosphen1, line 1: There would be no need for this import, remove it

Dimosphen1 (changes requested):

PR #998, opened 2024-08-08
Solution

Narberal90:


app/test_main.py
```diff
@@ -1 +1,39 @@
-# write your code here
+from typing import Callable
+
+import pytest
+from unittest import mock
+
+from app.main import can_access_google_page
+
+
```
oliasymak24, line 8: Use @pytest.parametrize

app/test_main.py
```diff
@@ -1 +1,39 @@
...
+    mock_has_internet.return_value = True
+    assert can_access_google_page("some_url") == "Not accessible"
+
+    mock_valid_google.return_value = True
+    mock_has_internet.return_value = False
+    assert can_access_google_page("some_url") == "Not accessible"
+
+    mock_valid_google.return_value = False
+    mock_has_internet.return_value = False
+    assert can_access_google_page("some_url") == "Not accessible"
```
oliasymak24, line 39: No need to write url for every case, u could make a constant from it

oliasymak24: Add an error message. It's good when the user understands what the problem is

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #1000, opened 2024-08-08
implement py-can-access-google-page

ValeriiaKyr:


app/test_main.py
```diff
@@ -1 +1,53 @@
...
+        mocked_valid_google_url: mock,
+        mocked_has_internet_connection: mock
+) -> None:
+    mocked_valid_google_url.return_value = True
+    mocked_has_internet_connection.return_value = False
+    assert can_access_google_page("http://www.google.com") == "Not accessible"
+
+
+@mock.patch("app.main.valid_google_url")
+@mock.patch("app.main.has_internet_connection")
```
oliasymak24, line 46: Imagine having a fixture and not writing those decorators again and again..

app/test_main.py
```diff
@@ -1 +1,53 @@
...
+def test_should_return_not_accessible_when_all_funcs_return_false(
+        mocked_valid_google_url: mock,
+        mocked_has_internet_connection: mock
+) -> None:
+    mocked_valid_google_url.return_value = False
+    mocked_has_internet_connection.return_value = False
+    assert can_access_google_page("http://www.google.com") == "Not accessible"
```
oliasymak24, line 53: No need to write url for every case, u could make a constant from it

app/test_main.py
```diff
@@ -1 +1,53 @@
-# write your code here
+from unittest import mock
+from app.main import can_access_google_page
+import pytest
+
+
```
oliasymak24, line 5: Add an error messages. It's good when the user understands what the problem is

app/test_main.py
```diff
@@ -1 +1,53 @@
-# write your code here
+from unittest import mock
+from app.main import can_access_google_page
+import pytest
+
```
oliasymak24, line 4: better to Use @pytest.parametrize

oliasymak24 (changes requested):

PR #1001, opened 2024-08-08
create mock test

panichqa:


app/test_main.py
```diff
@@ -1 +1,28 @@
...
+from app.main import can_access_google_page
+
+
+@mock.patch("app.main.valid_google_url")
+@mock.patch("app.main.has_internet_connection")
+def test_can_access_google_page(
+    mock_internet_connection: mock.Mock, mock_valid_url: mock.Mock
+) -> None:
+    mock_valid_url.return_value = True
+    mock_internet_connection.return_value = True
```
oliasymak24, line 11: better to Use @pytest.parametrize

app/test_main.py
```diff
@@ -1 +1,28 @@
...
+    assert can_access_google_page(
+        "https://www.invalid-url.com"
+    ) == "Not accessible"
+
+    mock_valid_url.return_value = False
+    mock_internet_connection.return_value = False
+    assert can_access_google_page(
+        "https://www.invalid-url.com"
+    ) == "Not accessible"
```
oliasymak24, line 28: Add an error messages. It's good when the user understands what the problem is

app/test_main.py
```diff
@@ -1 +1,28 @@
...
+    mock_valid_url.return_value = False
+    mock_internet_connection.return_value = True
+    assert can_access_google_page(
+        "https://www.invalid-url.com"
+    ) == "Not accessible"
+
+    mock_valid_url.return_value = False
+    mock_internet_connection.return_value = False
+    assert can_access_google_page(
+        "https://www.invalid-url.com"
```
oliasymak24, line 27: No need to write url for every case, u could make a constant from it

oliasymak24 (approved):

PR #1004, opened 2024-08-08
Solution

dpiuro:


app/test_main.py
```diff
@@ -1 +1,34 @@
...
+            assert can_access_google_page(
+                "https://www.google.com"
+            ) == "Accessible"
+
+
+def test_can_access_google_page_no_internet() -> None:
+    with patch("app.main.valid_google_url", return_value=True):
+        with patch("app.main.has_internet_connection", return_value=False):
+            assert can_access_google_page(
+                "https://www.google.com"
```
oliasymak24, line 17: No need to write url for every case, u could make a constant from it

app/test_main.py
```diff
@@ -1 +1,34 @@
...
+                "https://www.invalidurl.com"
+            ) == "Not accessible"
+
+
+def test_can_access_google_page_no_internet_and_invalid_url() -> None:
+    with patch("app.main.valid_google_url", return_value=False):
+        with patch("app.main.has_internet_connection", return_value=False):
+            assert can_access_google_page(
+                "https://www.invalidurl.com"
+            ) == "Not accessible"
```
oliasymak24, line 34: Add an error messages. It's good when the user understands what the problem is

app/test_main.py
```diff
@@ -1 +1,34 @@
-# write your code here
+from unittest.mock import patch
+from app.main import can_access_google_page
+
+
```
oliasymak24, line 4: better to use  @pytest.parametrize

oliasymak24 (changes requested):

app/test_main.py
dpiuro: Yes, it seems that I did not quite understand the task from the 1st time)

oliasymak24 (approved):

PR #1007, opened 2024-08-09
Solution of py-can-access-google-page

AndriiHamasa:


app/test_main.py
```diff
@@ -1 +1,36 @@
...
+        mocked_connection.return_value = True
+        mocked_valid_url.return_value = True
+
+        assert can_access_google_page("url") == "Accessible"
+
+
+def test_should_return_not_accessible() -> None:
+    with (
+        mock.patch("app.main.has_internet_connection") as mocked_connection,
+        mock.patch("app.main.valid_google_url") as mocked_valid_url
```
oliasymak24, line 31: Imagine having a fixture and not writing those mock.patch again and again..

app/test_main.py
```diff
@@ -1 +1,36 @@
...
+    ):
+        mocked_connection.return_value = True
+        mocked_valid_url.return_value = False
```
oliasymak24, line 34: U should add a test for False False too

app/test_main.py
```diff
@@ -1 +1,36 @@
-# write your code here
+from unittest import mock
+
+from app.main import can_access_google_page
+
+
```
oliasymak24, line 5: Use @pytest.parametrize

app/test_main.py
```diff
@@ -1 +1,36 @@
...
+def test_should_return_not_accessible() -> None:
+    with (
+        mock.patch("app.main.has_internet_connection") as mocked_connection,
+        mock.patch("app.main.valid_google_url") as mocked_valid_url
+    ):
+        mocked_connection.return_value = True
+        mocked_valid_url.return_value = False
+
+        assert can_access_google_page("url") == "Not accessible"
```
oliasymak24, line 36: No need to write url for every case, u could make a constant from it

oliasymak24 (changes requested):

app/test_main.py
```diff
@@ -1 +1,51 @@
...
+    mock_connection.assert_called_once()
+    mock_valid_url.assert_called_once_with(test_url)
+
+
+@pytest.mark.parametrize(
+    "connection_func_result,valid_url_func_result,expected_value",
+    [
+        (True, True, "Accessible"),
+        (True, False, "Not accessible"),
+        (False, False, "Not accessible")
```
oliasymak24, line 32: Also need a test for the case False, True

oliasymak24 (changes requested):

app/test_main.py
```diff
@@ -1 +1,53 @@
...
+from unittest import mock
+
+from app.main import can_access_google_page
+
+
+test_url = "url"
+
+
+@pytest.fixture()
+def mock_functions() -> tuple:
```
LiudmylaKulinchenko, line 11: Let's add more descriptive name here

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Looks good 🌸

PR #1009, opened 2024-08-09
Solution Politov

Alexandr-Politov:


app/test_main.py
```diff
@@ -1 +1,44 @@
-# write your code here
+import pytest
+from unittest import mock
+
+from app.main import can_access_google_page
+
+
```
oliasymak24, line 6: Use @pytest.parametrize

oliasymak24 (changes requested):

app/test_main.py
```diff
@@ -1 +1,43 @@
...
+    [
+        pytest.param(True, True, "Accessible",
+                     id="should return 'Accessible' when both url_validation"
+                        " and internet_connection are True"),
+        pytest.param(False, True, "Not accessible",
+                     id="should return 'Not accessible' "
+                        "when internet_connection is False"),
+        pytest.param(True, False, "Not accessible",
+                     id="should return 'Not accessible' "
+                        "when url_validation is False")
```
oliasymak24, line 31: U should add a test for False, False too

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #1010, opened 2024-08-09
Solution

Ivan-Shakhman:
Solution of task

### DIFF
--- a/app/test_main.py
+++ b/app/test_main.py
@@ -1,1 +1,26 @@
-# write your code here
+from unittest import mock
+from app.main import can_access_google_page
+
+
+def test_would_return_accessible_when_internet_and_valid_true():
+    with (mock.patch("app.main.valid_google_url", return_value=True),
+          mock.patch("app.main.has_internet_connection", return_value=True)):
+        assert (
+                can_access_google_page("1") == "Accessible"
+                ), f"Expected return 'Accessible' but was {can_access_google_page("1")}"
+
+
+def test_would_return_not_accessible_when_only_valid_true():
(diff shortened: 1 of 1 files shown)


app/test_main.py
```diff
@@ -1 +1,26 @@
-# write your code here
+from unittest import mock
+from app.main import can_access_google_page
+
+
```
oliasymak24, line 4: Use @pytest.parametrize

app/test_main.py
```diff
@@ -1 +1,26 @@
...
+    with (mock.patch("app.main.valid_google_url", return_value=False),
+          mock.patch("app.main.has_internet_connection", return_value=True)):
+        assert (
+                can_access_google_page("1") == "Not accessible"
+                ), f"Expected return 'Not accessible' but was {can_access_google_page("1")}"
+
+
+def test_would_return_not_accessible_when_only_connect_true():
+    with (mock.patch("app.main.valid_google_url", return_value=True),
+          mock.patch("app.main.has_internet_connection", return_value=False)):
```
oliasymak24, line 23: Imagine having a fixture and not writing those decorators again and again..

app/test_main.py
```diff
@@ -1 +1,26 @@
...
+        assert (
+                can_access_google_page("1") == "Not accessible"
+                ), f"Expected return 'Not accessible' but was {can_access_google_page("1")}"
```
oliasymak24, line 26: You must also add a test for case False, False

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #1014, opened 2024-08-10
Solution

efirshey:


NazarSheremet (changes requested):
Bring back the `@pytest.mark.parametrize` block to avoid duplicating the same function.

NazarSheremet (approved):
Good job!

PR #1016, opened 2024-08-10
Solution

olena4reunited:


app/test_main.py
```diff
@@ -1 +1,67 @@
...
+
+
+@pytest.mark.parametrize(
+    "connection_response, url_response, expected_result",
+    [
+        pytest.param(
+            True,
+            True,
+            "Accessible",
+            id="succesful access if google url is valid\
```
LiudmylaKulinchenko, line 33: try not to use `\` as a line breaker

app/test_main.py
```diff
@@ -1 +1,67 @@
...
+            True,
+            "Not accessible",
+            id="not accessible if url response is invalid"
+        ),
+        pytest.param(
+            False,
+            False,
+            "Not accessible",
+            id="not accessible if url response is invalid\
+                 and internet connection is unavailable"
```
LiudmylaKulinchenko, line 53: same here

app/test_main.py
```diff
@@ -1 +1,67 @@
-# write your code here
+import pytest
+from unittest import mock
+
+from app.main import can_access_google_page
+
+
+test_url = "url"
```
LiudmylaKulinchenko, line 7: It works, but let's make it look like url

LiudmylaKulinchenko (changes requested):

oliasymak24 (approved):

PR #1017, opened 2024-08-11
Solution

NazarSerdiuk1:


app/test_main.py
```diff
@@ -1 +1,28 @@
-# write your code here
+from unittest import mock
```
Dimosphen1, line 1: Add one blank line between different import groups to follow PEP-8

app/test_main.py
```diff
@@ -1 +1,28 @@
...
+        ]
+    )
+    @mock.patch("app.main.valid_google_url")
+    @mock.patch("app.main.has_internet_connection")
+    def test_can_access(self,
+                        mocked_google_url: mock,
+                        mocked_internet_connection: mock,
+                        valid_url: bool,
+                        connection: bool,
+                        expected_result: str) -> None:
```
Dimosphen1, line 24: Place each argument including `self` on a new line so that there would be no arguments on the same line with opening and closing parenthesis. It will help to improve the code style and readability

app/test_main.py
```diff
@@ -1 +1,28 @@
-# write your code here
+from unittest import mock
+import pytest
+
+from app.main import can_access_google_page
+
+
+class TestCanAccessGoogle:
```
Dimosphen1, line 7: There is no need to create `TestCanAccessGoogle` class, you can place the test function aside from it and remove the class

Dimosphen1 (changes requested):

app/test_main.py
```diff
@@ -1 +1,29 @@
-# write your code here
+import pytest
+
+from unittest import mock
```
Dimosphen1, line 3: Order your imports correctly according to [PEP-8](https://peps.python.org/pep-0008/#imports), `unittest` is a built-in library and `pytest` is third party

Dimosphen1 (changes requested):

PR #1019, opened 2024-08-11
'Solution'

ABaidin:


app/test_main.py
```diff
@@ -1 +1,39 @@
-# write your code here
+from unittest import mock
+
+from app.main import can_access_google_page
+
+
+@mock.patch("app.main.valid_google_url", return_value=True)
+@mock.patch("app.main.has_internet_connection", return_value=True)
```
oliasymak24, line 7: Imagine having a fixture and not writing those decorators again and again..

app/test_main.py
```diff
@@ -1 +1,39 @@
...
+def test_valid_url_and_connection_exists(
+        mock_has_internet_connection: bool,
+        mock_valid_google_url: bool
+) -> None:
+    assert can_access_google_page("http://www.google.com") == "Accessible"
```
oliasymak24, line 12: No need to write url for every case, u could make a constant from it

oliasymak24 (changes requested):

LiudmylaKulinchenko (approved):
Well done 🔥

PR #1020, opened 2024-08-12
Updated pull request

ihorutkin:


app/test_main.py
```diff
@@ -1 +1,36 @@
...
+from app.main import can_access_google_page
+
+
+@pytest.mark.parametrize(
+    "test_url,valid_google_url,has_internet_connection,expected",
+    [
+        ("https://www.google.com", True, True, "Accessible"),
+        ("https://www.google.com", True, False, "Not accessible"),
+        ("https://www.google.com", False, True, "Not accessible"),
+        ("https://www.google.com", False, False, "Not accessible"),
```
oliasymak24, line 13: No need to write url for every case, u could make a constant from it

oliasymak24 (changes requested):

app/test_main.py
```diff
@@ -1 +1,39 @@
 # write your code here
+import pytest
```
Oleksl888, line 2: check order of imports + separate types of imports with blank line

app/test_main.py
```diff
@@ -1 +1,39 @@
...
+from app.main import can_access_google_page
+
+
+url = "https://www.google.com"
+
+
+@pytest.mark.parametrize(
+    "test_url,valid_google_url,has_internet_connection,expected",
+    [
+        (url, True, True, "Accessible"),
```
Oleksl888, line 13: here we can don't have to pass url as a parameter since it's the same everywhere. You can pass it directly to test. It would make sense to parametrize it if it was different to every data set

app/test_main.py
```diff
@@ -1 +1,39 @@
...
+        "If we have valid google url and internet connection",
+        "If we have valid google url but don't have internet connection",
+        "If we don't have valid google url but have internet connection",
+        "If we don't have valid google url and internet connection",
+    ]
+)
+@patch("app.main.valid_google_url")
+@patch("app.main.has_internet_connection")
+def test_correct(
+        mock_valid_url: bool,
```
Oleksl888, line 28: wrong type for mock function object

app/test_main.py
```diff
@@ -1 +1,39 @@
...
+    ids=[
+        "If we have valid google url and internet connection",
+        "If we have valid google url but don't have internet connection",
+        "If we don't have valid google url but have internet connection",
+        "If we don't have valid google url and internet connection",
+    ]
+)
+@patch("app.main.valid_google_url")
+@patch("app.main.has_internet_connection")
+def test_correct(
```
Oleksl888, line 27: can you think of a more appropriate test name?

app/test_main.py
```diff
@@ -1 +1,39 @@
...
+        mock_has_internet: bool,
+        test_url: str,
+        valid_google_url: bool,
+        has_internet_connection: bool,
+        expected: str
+) -> None:
+    mock_valid_url.return_value = valid_google_url
+    mock_has_internet.return_value = has_internet_connection
+
+    result = can_access_google_page(test_url)
```
Oleksl888, line 38: why do you need to store result in a variable instead of calling function in assert directly

app/test_main.py
```diff
@@ -1 +1,39 @@
...
+    assert result == expected
```
Oleksl888, line 39: you should also assert that mocked functions has been called

Oleksl888 (changes requested):

app/test_main.py
```diff
@@ -1 +1,36 @@
 # write your code here
```
LiudmylaKulinchenko, line 1: All good, but let's remove unnecessary comment 🤠

LiudmylaKulinchenko (changes requested):

app/test_main.py
```diff
@@ -1 +1,35 @@
...
+@pytest.mark.parametrize(
+    "valid_google_url,has_internet_connection,expected",
+    [
+        (True, True, "Accessible"),
+        (True, False, "Not accessible"),
+        (False, True, "Not accessible"),
+        (False, False, "Not accessible"),
+    ],
+    ids=[
+        "If we have valid google url and internet connection",
```
Oleksl888, line 17: You can have more descfriptive messages:
```
        "Accessible If we have valid google url and internet connection",
        "Not accessible If we have valid google url but don't have internet connection",

File: app/test_main.py [hunk 10]
```diff
@@ -1 +1,35 @@
...
+        mock_valid_url: MagicMock,
+        mock_has_internet: MagicMock,
+        valid_google_url: bool,
+        has_internet_connection: bool,
+        expected: str
+) -> None:
+    mock_valid_url.return_value = valid_google_url
+    mock_has_internet.return_value = has_internet_connection
+
+    assert can_access_google_page("https://www.google.com") == expected
```
Oleksl888, line 35: can you also add asserts that mock functions has been called?

Oleksl888 (changes requested):

Oleksl888 (approved):
Well done!

PR #1024, opened 2024-08-18
Solution

zazmarga:


vasylhnatiuk (changes requested):
you should check all options - True, False and False, False as well

PR #1033, opened 2024-08-25
'Solution'

PVS1905:


vasylhnatiuk (commented):
add test for False, False. and it would be better to use pytest.parametrize

vasylhnatiuk (changes requested):
add test for False, False. and it would be better to use pytest.parametrize

PR #1035, opened 2024-08-27
solution

Kulibchenko:


NazarSheremet (changes requested):
Bring back the `@pytest.mark.parametrize` block to avoid duplicating the same function.

YuriiSmolii (approved):
LGTM

PR #1038, opened 2024-08-29
Solution

Serhii-Chubur:


app/test_main.py
```diff
@@ -1 +1,39 @@
-# write your code here
+import pytest
+from app.main import can_access_google_page
+from unittest.mock import patch, MagicMock
```
LLkaia, line 3: Add blank line between different kind of imports

app/test_main.py
```diff
@@ -1 +1,39 @@
...
+
+
+@pytest.fixture()
+def url() -> str:
+    url = "https://Chubur/py-can-access-google-page"
+    return url
```
LLkaia, line 9:
```suggestion
    return "https://Chubur/py-can-access-google-page"
```

LLkaia (changes requested):
Additionally, move return value from test to `patch` decorator.

vasylhnatiuk (changes requested):
add test for False, False as well

app/test_main.py
```diff
@@ -1 +1,41 @@
...
+                                    url: str) -> None:
+    assert can_access_google_page(url) == "Not accessible"
+
+
+@patch("app.main.has_internet_connection", return_value=False)
+@patch("app.main.valid_google_url", return_value=False)
+def test_invalid_url_and_no_conection(mock_valid_google_url: MagicMock,
+                                      mock_has_internet_connection: MagicMock,
+                                      url: str) -> None:
+    assert can_access_google_page(url) == "Not accessible"
```
vsmutok, line 41: please use `pytest.mark.parametrize`

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #1039, opened 2024-08-30
solution

nrjss67:


vasylhnatiuk (changes requested):
add tests if  `valid_url.return_value = False` or `internet_connection.return_value = False`

vsmutok (approved):
Great!

PR #1043, opened 2024-08-31
Solution

PavelTym:


app/test_main.py
```diff
@@ -1 +1,55 @@
-# write your code here
+import unittest
+from unittest.mock import patch, MagicMock
+
+from app.main import can_access_google_page
+
+
+class TestCanAccessGooglePage(unittest.TestCase):
+
```
pavlopro, line 8: Please use @pytest.mark.parametrize decorator to avoid code duplication

pavlopro (changes requested):

pavlopro (approved):
GJ!

PR #1049, opened 2024-09-03
Solution

juliazelena:


app/test_main.py
```diff
@@ -1 +1,40 @@
-# write your code here
+from unittest import mock
+from typing import Any
+import pytest
```
pavlopro, line 3: Add blank line between diffrent import groups

app/test_main.py
```diff
@@ -1 +1,40 @@
...
+def mocked_valid_google_url() -> None:
+    with (mock.patch("app.main.valid_google_url") as mock_valid_google_url):
+        yield mock_valid_google_url
+
+
+@pytest.fixture()
+def mocked_has_internet_connection() -> None:
+    with (mock.patch("app.main.has_internet_connection") as
+          mock_has_internet_connection):
+        yield mock_has_internet_connection
```
pavlopro, line 18: Instead of pytest.fixtute apply pach decorator

```
@mock.patch("app.main.valid_google_url")
@mock.patch("app.main.has_internet_connection")
```

pavlopro (changes requested):

app/test_main.py
```diff
@@ -1 +1,31 @@
-# write your code here
+from unittest import mock
+
```
pavlopro, line 2: no need a blank line between 
```
from unittest import mock
from typing import Any
```

They both related to the same group

app/test_main.py
```diff
@@ -1 +1,31 @@
...
+        (True, True, "Accessible"),
+        (True, False, "Not accessible"),
+        (False, True, "Not accessible"),
+        (False, False, "Not accessible")
+    ]
+)
+@mock.patch("app.main.valid_google_url")
+@mock.patch("app.main.has_internet_connection")
+def test_can_access_google_page(
+        mocked_valid_google_url: Any,
```
pavlopro, line 22: Correct annotation
should be mock.MagicMock

pavlopro (changes requested):

PR #1051, opened 2024-09-08
can access google page

AndreyHMkr:


app/test_main.py
```diff
@@ -1 +1,49 @@
-# write your code here
+import pytest
+from unittest import mock
+from unittest.mock import MagicMock
+from app.main import can_access_google_page
+
```
pavlopro, line 5: please use:

```
@pytest.mark.parametrize
@mock.patch
```

pavlopro (changes requested):

app/test_main.py
```diff
@@ -1 +1,29 @@
...
+from unittest import mock
+from app.main import can_access_google_page
+
+
+@pytest.mark.parametrize(
+    "test_link,test_connection,expected_result",
+    [
+        (True, True, "Accessible"),
+        (True, False, "Not accessible"),
+        (False, True, "Not accessible")
```
pavlopro, line 13: Please add one more case with (False, False)

app/test_main.py
```diff
@@ -1 +1,29 @@
...
+    ]
+
+)
+@mock.patch("app.main.has_internet_connection")
+@mock.patch("app.main.valid_google_url")
+def test_connection(
+        mock_valid_url: Any,
+        mock_connection: Any,
```
pavlopro, line 21: Please correct return annotation from `Any` to `mock.MagicMock`

app/test_main.py
```diff
@@ -1 +1,29 @@
...
+        test_connection: bool,
+        test_link: str,
+        expected_result: str
+) -> None:
+    mock_valid_url.return_value = test_link
+    mock_connection.return_value = test_connection
+    result = can_access_google_page(mock_valid_url)
```
pavlopro, line 28: No need to store result into variable. Just directly compare with expected result

pavlopro (changes requested):

app/test_main.py
```diff
@@ -1 +1,30 @@
-# write your code here
+from typing import Any
+
+import pytest
+from unittest import mock
+from app.main import can_access_google_page
```
pavlopro, line 5: Add blank line between different import groups

pavlopro (changes requested):

PR #1054, opened 2024-09-11
Solution

bdnnk:


app/test_main.py
```diff
@@ -1 +1,25 @@
...
+        (False, False, "Not accessible")
+    ]
+)
+def test_can_access_google_page(
+        connection: bool,
+        url_valid: bool,
+        expected_result: str
+) -> None:
+    with (patch("app.main.has_internet_connection", return_value=connection)):
+        with patch("app.main.valid_google_url", return_value=url_valid):
```
Dimosphen1, line 22: Do not creatre two nested context managers, add two mocks as decorators and pass them as arguments instead

app/test_main.py
```diff
@@ -1 +1,25 @@
-# write your code here
+import pytest
+from unittest.mock import patch
+
+from app.main import can_access_google_page
```
Dimosphen1, line 4: Order and group your imports to follow PEP-8

Dimosphen1 (changes requested):

app/test_main.py
```diff
@@ -1 +1,29 @@
...
+        (True, False, "Not accessible"),
+        (False, True, "Not accessible"),
+        (False, False, "Not accessible")
+    ]
+)
+@mock.patch("app.main.has_internet_connection")
+@mock.patch("app.main.valid_google_url")
+def test_can_access_google_page(
+        mock_valid_google_url: Callable,
+        mock_has_internet_connection: Callable,
```
pavlopro, line 22: Correct annotation to `mock.MagicMock`

pavlopro (changes requested):

app/test_main.py
```diff
@@ -1 +1,29 @@
-# write your code here
+from unittest import mock
+from unittest.mock import MagicMock
```
pavlopro, line 2: You already have imported mock. Just use it instead of importing MagicMock

pavlopro (changes requested):

PR #1066, opened 2024-09-27
Solution

p1tonidze:


NazarSheremet (changes requested):
Need  to use `@pytest.mark.parametrize` to not dublicate the same block 3 times

NazarSheremet (approved):
Good job!

PR #1072, opened 2024-10-01
Solution

watergor:


vasylhnatiuk (changes requested):
you have to check different results not only True, True

vasylhnatiuk (approved):

PR #1073, opened 2024-10-02
solution

dmytrik:


app/test_main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+from unittest import mock
+
+
+from app.main import can_access_google_page
+
+
+@mock.patch("app.main.has_internet_connection")
+def test_has_been_has_internet_connection_called(
```
AnyoneClown, line 8: Let's add better test name. Example `test_internet_connection_called`

Rename second test the same way

app/test_main.py
```diff
@@ -1 +1,20 @@
...
+        m_has_internet_connection: object
+) -> None:
+    m_has_internet_connection.return_value = True
+    can_access_google_page("https://www.google.com/")
```
AnyoneClown, line 12: Python doesn't really have any constants, but you can put the url that we use a lot at the beginning of the file as a constant.

Example:
```
GOOGLE_URL = "https://www.google.com/"
```

```suggestion
    can_access_google_page(GOOGLE_URL)
```

AnyoneClown (changes requested):

AnyoneClown (approved):
Good, but check all comments carefully. You forgot to change second test name

PR #1077, opened 2024-10-07
Solution

KKvladys:


app/test_main.py
```diff
@@ -1 +1,37 @@
...
+        mock_has_internet: MagicMock,
+        url: str,
+        has_internet_connection: bool,
+        valid_google_url: bool,
+        can_access: str
+) -> None:
+    mock_has_internet.return_value = has_internet_connection
+    mock_valid_url.return_value = valid_google_url
+
+    assert can_access_google_page(url) == can_access
```
Mxbely, line 37: you don't check inner functions, you just mock and set result them.
`has_internet_connection` and `valid_google_url` are collable, i think, you need make a test for check call inner functions

Mxbely (approved):
I approve it, but you need to think about my comments

app/test_main.py
```diff
@@ -1 +1,37 @@
...
+        pytest.param(
+            "https://www.google.com", False, True, "Not accessible",
+            id="Not accessible if no internet connection"
+        ),
+        pytest.param(
+            "https://www.youtube.com", True, False, "Not accessible",
+            id="Not accessible if no valid google url"
+        )
+    ]
+)
```
AnyoneClown, line 23: Add one more test for `False` `False`(No internet + invalid URL)

app/test_main.py
AnyoneClown: Right, we can use `assert_called_once()` for that

AnyoneClown (changes requested):

AnyoneClown (approved):
Good job!

PR #1087, opened 2024-10-15
can-access-google-page solution

94nj111:


app/test_main.py
```diff
@@ -1 +1,77 @@
-# write your code here
+from typing import Callable
+
+import pytest
+from unittest import mock
```
vladyslav-tmf, line 4: Your code is great!
The only thing that needs to be fixed is the order of imports.
You correctly import `Callable` first, because it is a built-in module, but `unittest` is also built-in module, so it should also be placed after typing.
```suggestion
from typing import Callable
from unittest import mock

import pytest
```

vladyslav-tmf (approved):
Good job!

PR #1090, opened 2024-10-16
 Solution-py-can-access-google-page

skyfoxwork:


app/test_main.py
```diff
@@ -1 +1,16 @@
...
+@mock.patch("app.main.valid_google_url")
+def test_valid_google_url_exist(mocked_url: Any) -> None:
+    can_access_google_page("http://google.com")
+    mocked_url.assert_called_once()
+
+
+@mock.patch("app.main.has_internet_connection")
+def test_has_internet_connection_exist(mocked_connection: Any) -> None:
+    can_access_google_page("http://google.com")
+    mocked_connection.assert_called_once()
```
Atikiho, line 16: You don't have limits to mock functions in test, so you can implement it in one test. Like this:
```
@mock.patch()
@mock.patch()
def your_test()
```
You also should annotate mocked functions like mock.MagicMock, not Any

Atikiho (changes requested):

app/test_main.py
skyfoxwork: Thank you for the good tip !

PR #1093, opened 2024-10-19
'Solution'

spa7id:


Mxbely (changes requested):
1. you can use pytest parametrize with one function instead of three tests functions
2. if you mocking functions, you can check they for callable (maybe it's options, but it will be better)

PR #1094, opened 2024-10-19
Solution

Den-k0:


app/test_main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+from unittest import mock
```
VladyslavBon, line 1: Better use 1 space between this two types of imports, but in general code looks good

VladyslavBon (approved):

PR #1108, opened 2024-10-21
solution

SerhiiK05:


app/test_main.py
```diff
@@ -1 +1,47 @@
-# write your code here
+from unittest import mock
+
+from app.main import (can_access_google_page)
+
+
+@mock.patch("app.main.valid_google_url")
+@mock.patch("app.main.has_internet_connection")
+def test_valid_url_and_connection_exists(mocked_valid_google_url: bool,
+                                         mocked_has_internet_connection: bool
```
sumseer, line 9: mocked arguments are not bool, but MagiMock objects. correct this everywhere

sumseer (changes requested):

sumseer (approved):

PR #1113, opened 2024-10-21
Solution

sofiiasavkova:


dxrrkwm (changes requested):
i would do it using the context manager, but using this approach i think you should patch it on classm not every method

PR #1121, opened 2024-10-24
Solution

apmaz:


app/test_main.py
```diff
@@ -1 +1,29 @@
-# write your code here
+from unittest import mock
+
+from app.main import can_access_google_page
+
+
+@mock.patch("app.main.valid_google_url")
+@mock.patch("app.main.has_internet_connection")
+def test_access_google_page_with_true_false(
```
pavlopro, line 8: Please use parametrize decorator and add cases where valid_url and connection are both False and True

pavlopro (changes requested):

pavlopro (approved):
Good Job!

PR #1126, opened 2024-10-29
Solution

serg121192:


app/test_main.py
```diff
@@ -1 +1,27 @@
...
+url = "https://google.com"
+
+
+@mock.patch("app.main.valid_google_url")
+@mock.patch("app.main.has_internet_connection")
+def test_can_access_google_page_with_mocked_different_dependencies(
+        mocked_url: MagicMock,
+        mocked_connection: MagicMock
+) -> None:
+    possible_variants = [
```
pavlopro, line 16: Use parametrize decorator instead

app/test_main.py
```diff
@@ -1 +1,58 @@
-# write your code here
+from unittest import mock
+from unittest.mock import MagicMock
+
+import pytest
+
+from app.main import can_access_google_page
+
+
+url = "https://google.com"
```
Dimosphen1, line 9: It is better to use upper case here
```suggestion
TEST_URL = "https://google.com"
```

app/test_main.py
```diff
@@ -1 +1,58 @@
...
+
+@pytest.mark.parametrize(
+    "mocked_url_return_value,"
+    "mocked_connection_return_value,"
```
Dimosphen1, line 13: It may be no need to include these values here since you use decorators

Dimosphen1 (changes requested):
Several changes were requested. Please, make sure that all checks passed

PR #1180, opened 2024-12-17
'Solution'

dottfmar:


app/test_main.py
```diff
@@ -1 +1,23 @@
-# write your code here
+import app.main
+from app.main import can_access_google_page
+from unittest import mock
+
+
+def test_access_google_page_without_internet_connection() -> None:
```
taraskhom, line 6: You can also make one function and use pytest to set return_value for two required functions and excpected_value for checking the behavior

taraskhom (approved):

PR #1189, opened 2024-12-19
Solution

maxakim0v:


4ikkamony (approved):
Nice!

Little comment, though:
Consider making a separate test for each assertion, so that it is clear which one has failed.
Currently, if the first one fail, you will not know what other assertions' results are

PR #1190, opened 2024-12-19
Access google page tests

4ikkamony:


valerii-kashpur (approved):

L1nk3rrr (approved):
Well done! You show a full understanding of Pytest here!
But if you do this with a unit test mock patch, it can look more pretty, but it's a matter of taste. Good job!

4ikkamony:
@L1nk3rrr Thanks! I wasn't sure if this approach was fine, I had like a few other versions and pushed this as the least ugly-looking one

PR #1207, opened 2024-12-21
 Create task Can access google page

ihorhalyskiy:


Professor-Douel (approved):

L1nk3rrr (changes requested):
Hi! What if you try to patch has_internet_connection and valid_google_url in once?
Also you can have such combinations for testing
```
has_internet_connection                  can_access_google_page
False                                    False
False                                    True
True                                     True
True                                     False
```

L1nk3rrr (changes requested):
It's a bad idea to have such variables and do all tests in one. Your assert message duplicates your variable values, not giving any information to the user.

Please make separate functions with test cases that you have in my previous message

L1nk3rrr (approved):
Well done!

PR #1208, opened 2024-12-21
Solution

Professor-Douel:


app/test_main.py
```diff
@@ -1 +1,61 @@
-# write your code here
+import unittest
```
L1nk3rrr, line 1: from unittest you need only TestCase, not a full module

L1nk3rrr (changes requested):
Hi, well done!
Minor points - invalid URL makes no sense here. Because the function is accessible for Google, so your naming like `invalid_url` doesn't describe any functionality that will affect the result.

rename your function to fully describe what tests do:
```
test_accessible
test_no_internet
test_invalid_url
test_no_internet_and_invalid_url
```

app/test_main.py
```diff
@@ -1 +1,61 @@
...
+from unittest.mock import patch, Mock
```
L1nk3rrr, line 2: It will be better to do smth like this
`from unittest import mock, TestCase`

And then use `mock.patch`, because a lot of modules can have such name of function

L1nk3rrr (changes requested):
Hi, well done!
Minor points - invalid URL makes no sense here. Because the function is accessible for Google, so your naming like `invalid_url` doesn't describe any functionality that will affect the result.

rename your function to fully describe what tests do:
```
test_accessible
test_no_internet
test_invalid_url
test_no_internet_and_invalid_url
```

L1nk3rrr (approved):
Well done!!!

PR #1213, opened 2024-12-22
solution_v.1.0

yelizaveta-zh:


app/test_main.py
```diff
@@ -1 +1,54 @@
...
+@patch("app.main.has_internet_connection")
+def test_can_access_google_page_accessible(
+        mock_internet: patch,
+        mock_valid_url: patch
+) -> None:
+    mock_internet.return_value = True
+    mock_valid_url.return_value = True
+
+    result = can_access_google_page("https://www.google.com")
+    assert result == "Accessible"
```
Nattalli, line 15: No need to create new variable

```suggestion
    assert can_access_google_page("https://www.google.com") == "Accessible"
```

Nattalli: also add message with error description, for example:

```suggestion
    assert can_access_google_page("www.google.com/") == "Accessible", (
        "Should be accessible when URL is valid and there is an internet connection."
    )
```

Nattalli: the same everywhere

Nattalli (changes requested):

L1nk3rrr (approved):
Well done!

PR #1214, opened 2024-12-23
Solution

bodiakof:


requirements.txt
```diff
@@ -6,3 +6,5 @@ flake8-variables-names==0.0.5
 pep8-naming==0.13.2
 requests==2.27.1
 pytest-xdist==2.5.0
+
+pytest-mock~=3.14.0
```
L1nk3rrr, line 10: do you really need an external lib?

L1nk3rrr (changes requested):
You have mock at unittest, you can combine that libs, for such task you don't need external libs.

requirements.txt
bodiakof: i just got this bright idea to use everything from pytest, since i use pytest for testing and not unittest :))

L1nk3rrr (approved):
Wekk done!

PR #1223, opened 2025-01-03
Solution

MarianaAB19:


app/test_main.py
```diff
@@ -1 +1,27 @@
...
+        ("https://www.gooogle.com.ua", False, False, "Not accessible"),
+    ]
+)
+def test_valid_url_and_connection_exists(
+        url: str,
+        valid_url: bool,
+        internet: bool,
+        result: str
+) -> None:
+    with (mock.patch("app.main.valid_google_url")
```
pavlopro, line 21: please use mock.path as decorator to make it more readable

pavlopro (changes requested):

vsmutok (approved):
Looks Good!

PR #1276, opened 2025-02-07
Solution

Baranotik15:


kram3ko (approved):
Good Job!.  As i remember need to check: can_access_google_page =) extra tests guess fine.

app/test_main.py
```diff
@@ -1 +1,68 @@
...
+        ) == "Not accessible"
+
+
+def test_can_access_google_page_invalid_url() -> None:
+    with (patch("app.main.has_internet_connection") as mock_has_internet,
+          patch("app.main.valid_google_url") as mock_valid_url):
+        mock_has_internet.return_value = True
+        mock_valid_url.return_value = False
+        assert can_access_google_page(
+            "https://www.google.com/"
```
oliasymak24, line 67: No need to write url for every case, u could make a constant from it

app/test_main.py
```diff
@@ -1 +1,68 @@
-# write your code here
+from unittest.mock import patch, Mock, MagicMock
+from .main import (can_access_google_page,
+                   has_internet_connection,
+                   valid_google_url)
+
+
+def test_valid_google_url() -> None:
```
oliasymak24, line 7: You can use `parametrize` decorator to avoid duplicate code

oliasymak24 (changes requested):

PR #1290, opened 2025-02-17
'Solution'

JohnyJDX:


app/test_main.py
```diff
@@ -1 +1,29 @@
...
+from app.main import can_access_google_page
+
+
+@pytest.mark.parametrize(
+    "mock_valid_url,mock_internet_connection, expected_result",
+    [
+        (True, True, "Accessible"),
+        (True, False, "Not accessible"),
+        (False, True, "Not accessible"),
+        (False, False, "Not accessible"),
```
Nattalli, line 12: okay but add description of the errors

Nattalli: you can add `id` parameter, for example:
```
        (False, False, "Not accessible", "should return 'Not accessible' when both functions return `False`"),

Nattalli (changes requested):

PR #1311, opened 2025-03-06
solution

cemen2009:


app/test_main.py
```diff
@@ -1 +1,46 @@
-# write your code here
+from pytest_mock import MockerFixture
```
HappyDen08, line 1: pytest_mock?

HappyDen08 (commented):
REplace pytest_mock? U mean unittest mock?

HappyDen08 (changes requested):
replace pytest_mock on unettest mock

PR #1329, opened 2025-03-09
task solution

yarms1:


ntkachenko21 (approved):
you have a lot of repeated lines of code, but it works so good job

PR #1360, opened 2025-03-23
Solution

BuryiOstap:


app/test_main.py
```diff
@@ -1 +1,16 @@
...
+
+@mock.patch("app.main.valid_google_url")
+@mock.patch("app.main.has_internet_connection")
+def test_valid_url_and_connection_exists(
+        mocked_valid_google_url: mock.MagicMock,
+        mocked_has_internet_connection: mock.MagicMock
+) -> None:
+    can_access_google_page(mocked_valid_google_url)
+    mocked_valid_google_url.assert_called_once()
+    mocked_has_internet_connection.assert_called_once()
```
vsmutok, line 14: The current test only checks if the mocked functions are called. It does not assert the return value of the `can_access_google_page` function based on different return values of the mocked functions. You need to add assertions to verify the logic of the `can_access_google_page` function.

Consider adding test cases to check the following:

* Both `has_internet_connection` and `valid_google_url` return `True`.
* `has_internet_connection` returns `False` and valid_google_url returns `True`.
* `has_internet_connection` returns `True` and valid_google_url returns `False`.
* Both `has_internet_connection` and `valid_google_url` return `False`.

vsmutok (changes requested):

app/test_main.py
```diff
@@ -1 +1,25 @@
...
+    ]
+)
+def test_valid_url_and_connection_exists(
+    url: str,
+    valid_google_url: str,
+    has_internet_connection: bool,
+    result: str
+) -> None:
+    with mock.patch("app.main.valid_google_url",
+                    return_value=valid_google_url), \
```
vsmutok, line 22: better to use `()` then `\`

vsmutok (approved):

PR #1469, opened 2025-07-10
Implemented solution.

Otst0y:


Orixara (approved):
interesting way to solve, I like it

PR #1470, opened 2025-07-10
my-solution

Orixara:


SerhiiBasok (approved):
Good job, I like how visually your code looks.👍🏻 You can also use the pytest decorator in this task, which will allow us not to rewrite the tests, and also add an id to each test.

PR #1481, opened 2025-07-13
Solution

yagginov:


Gleg0 (approved):
Well done.

yagginov:
@mateacademy-ai-mentor 
You're mistaken — the order of the arguments in the test function is actually **correct**. When using multiple `@mock.patch` decorators, they are applied **from the bottom up**, meaning:

```python
@mock.patch("app.main.valid_google_url")  # applied second
@mock.patch("app.main.has_internet_connection")  # applied first
```

This results in the first argument in the function being the mock for `has_internet_connection`, and the second — for `valid_google_url`, which exactly matches:

```python
def test_can_access_google_page(
    mocked_has_internet_connection,
    mocked_valid_google_url,
    ...
)
```

So no, there’s nothing to swap — the function signature already aligns perfectly with the decorator order.

PR #1497, opened 2025-07-16
Solution

Bebaaaaa:


bohdankolisnyk04 (changes requested):
The current test code is quite tangled and has some structural issues that make it hard to read and maintain. There are nested function definitions inside tests, inconsistent use of patching and monkeypatching, and even calls to pytest inside the tests themselves, which is unconventional. It would be much clearer and more reliable to separate each test properly, avoid nested functions, stick to one mocking method (preferably using decorators), and remove any test runner calls from the test functions. Cleaning up the formatting and organizing the tests this way will make the suite easier to work with and more robust.

PR #1679, opened 2026-01-03
Solution

tarasmosiichuk01-ship-it:


app/test_main.py
```diff
@@ -1 +1,71 @@
...
+        app.main,
+        "valid_google_url",
+        lambda url: True
+    )
+    monkeypatch.setattr(
+        app.main,
+        "has_internet_connection",
+        lambda: False
+    )
+    assert can_access_google_page(url=adress_url) == "Not accessible"
```
vsmutok: Try using **`pytest.mark.parametrize`** to avoid duplication here.

vsmutok (changes requested):

vsmutok (changes requested):
You should remove the four individual test functions (`test_valid_url_and_connection_exists`, `test_valid_url_but_no_connection`, `test_invalid_url_but_connection_exists`, `test_invalid_url_and_no_connection`) because they duplicate the logic already covered by the single parametrized test. The parametrized test with `@pytest.mark.parametrize` makes the code concise, easier to maintain, and avoids unnecessary repetition.

vsmutok (approved):
Great!

PR #1753, opened 2026-03-03
Written tests for function

SerhiiHryniuk:


Reverad (changes requested):
use parametrize from pytest framework to keep the code DRY

PR #1791, opened 2026-04-02
develop v1

LayMiDD:


app/test_main.py
```diff
@@ -1 +1,44 @@
-# write your code here
+from unittest import mock
+from app.main import can_access_google_page
+
+
+def test_access_when_url_and_connection_are_valid() -> None:
```
vsmutok: Use `pytest.mark.parametrize` to avoid code duplication.

vsmutok (changes requested):

app/test_main.py
```diff
@@ -1 +1,28 @@
...
+
+from app.main import can_access_google_page
+
+
+@pytest.mark.parametrize(
+    "internet_connection, valid_url, expected_result",
+    [
+        (True, True, "Accessible"),
+        (True, False, "Not accessible"),
+        (False, True, "Not accessible"),
```
vsmutok, line 13: add test for `False` and `False` case

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #1906, opened 2026-06-28
Develop

andriideneha-dotcom:


app/test_main.py
```diff
@@ -1 +1,30 @@
...
+        with patch("app.main.has_internet_connection", return_value=True):
+            result = can_access_google_page("https://google.com")
+            assert result == "Accessible"
+
+
+def test_invalid_url() -> None:
+    with patch("app.main.valid_google_url", return_value=False):
+        with patch("app.main.has_internet_connection", return_value=True):
+            result = can_access_google_page("https://invalid.com")
+            assert result == "Not accessible"
```
vsmutok: Try using `@pytest.mark.parametrize` here to avoid code duplication.

vsmutok (changes requested):

app/test_main.py
andriideneha-dotcom: Thanks, that's good advice, for some reason, I hadn't thought of that

PR #1911, opened 2026-06-30
feat: Solution

sashokpilipok7:


app/test_main.py
```diff
@@ -1 +1,40 @@
...
+@mock.patch("app.main.valid_google_url")
+def test_can_access_google_page_wih_no_connection(
+        mocked_has_connection: mock.MagicMock,
+        mocked_valid_url: mock.MagicMock) -> None:
+    mocked_valid_url.return_value = True
+    mocked_has_connection.return_value = False
+    assert can_access_google_page("https://data.rocks") == "Not accessible"
+
+    mocked_valid_url.assert_called()
+    mocked_has_connection.assert_called()
```
vsmutok, line 29: Try using `@pytest.mark.parametrize` here to avoid code duplication.

vsmutok (changes requested):

